PR #26, opened 2017-08-31
Add support for to choose the mount point of the app

vdel:
Vesta services are mounted into various locations. Instead of changing all the docker-compose files, one could change the mount point (`/app` by default) in dmake.


deepomatic/dmake/deepobuild.py
```diff
@@ -85,33 +84,25 @@ class EnvBranchSerializer(YAML2PipelineSerializer):
...
-            for k, v in additional_variables.items():
-                 variables[k] = v
+                cmd = ['source ' + self.source + ' && env']
+                env = []
+                for v in common.run_shell_command(cmd).split('\n'):
+                    v = v.split('=')
+                    variable = v[0]
+                    value = '='.join(v[1:])
+                    env.append(variable + '=' + common.wrap_cmd(value))
+                env = ' '.join(env)
```
thomas-riccardi: Not needed to cache `'source ' + self.source` result; it should not be that costly.
Furthermore, the current way is not safe: `variable + '=' + common.wrap_cmd(value)` is not a valid bash serialization.
Example error:
```
deepomatic.dmake.common.ShellError: bash: BASH_FUNC_declare_grep_exclude_arguments%%=() {  local __grep_exclude_arguments=();: command not found
bash: [[: command not found
bash: =: command not found
bash: [[: command not found
bash: =: command not found
```

I'll inline  `'source ' + self.source`.

deepomatic/dmake/deepobuild.py
```diff
@@ -85,33 +84,25 @@ class EnvBranchSerializer(YAML2PipelineSerializer):
...
+            else:
+                env = ""
 
-            cmd.append(delimitor)
-            for value in variables.values():
-                cmd.append('echo %s' % common.wrap_cmd(value))
-            cmd.append(delimitor)
+            variables = dict(self.variables.items() + additional_variables.items())
+            for var, value in variables.items():
+                replaced_variables[var] = common.run_shell_command(env + ' && echo %s' % common.wrap_cmd(value)).replace('\n', '')
```
thomas-riccardi: `replace('\n', '')` is only OK for json values, which is the use-case now.
It's hard to do better because `docker run --env-file` does *not* support multiline values (nor `Dockerfile`) (see https://github.com/moby/moby/issues/12997 and https://github.com/moby/moby/issues/19906).
I'll add a comment.

thomas-riccardi:
To do:
- [x] test shell, build, run
- [ ] test aws deployment

deepomatic/dmake/serializer.py
```diff
@@ -40,7 +40,12 @@ def __init__(self,
         if not isinstance(data_type, list):
             data_type = [data_type]
         for t in data_type:
-            assert(isinstance(t, YAML2PipelineSerializer) or isinstance(t, FieldSerializer) or t in self.allowed_types)
+            if not (isinstance(t, YAML2PipelineSerializer) or isinstance(t, FieldSerializer) or t in self.allowed_types):
+                if not isinstance(t, tuple):
+                    raise Exception('Unknown type: %s' % str(t))
+                for tt in t:
+                    if not tt in self.allowed_types:
```
thomas-riccardi: `E713 test for membership should be 'not in'`

deepomatic/dmake/serializer.py
```diff
@@ -40,7 +40,12 @@ def __init__(self,
...
+                        raise Exception('Unknown type: %s' % str(tt))
```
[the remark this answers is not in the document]
vdel: nicely spoted: it was the reminiscence of a hacky attempt for doing something, that I gave up. I forgot to undo this change as well.

PR #68, opened 2017-11-03
Fix dmake stop

thomas-riccardi:
`dmake stop` should stop containers started by `dmake start`.
It previously failed because it searched wrong container names.

The run_daemon is called with empty name, and auto generated names have
`repo.branch.build` as prefix: using that to stop containers.


thomas-riccardi:
In fact stop now stops too much: it stops all containers, not just the one service specified in cmdline, but it's less bad than before.

PR #81, opened 2017-11-27
Renaming 'docker_links_names' into 'needed_links'

vdel:
The `docker_links_names` is actually used by `test` and `run` commands.

For consistency, I moved `docker_links_names` on the same level as `needed_services` and renamed its `needed_links`.

Backward compatibility is maintained thanks to a new 'migrations' notion: yaml can be patched and re-exported. It's important to name migrations by starting with '0xyx' to order their dependancy.

As I added a new python librairie, I patched the tests to install a virtual environment. 
**TODO: It does not cover any python script in utils as the `load('Dmakefile')` step uses default Jenkins python environment.**

I also added Python3 checks on the fly.

### DIFF
--- a/deepomatic/dmake/common.py
+++ b/deepomatic/dmake/common.py
@@ -13,9 +13,9 @@ logger.addHandler(logging.StreamHandler())
 ###############################################################################
 
 if sys.version_info >= (3,0):
-    from deepomatic.dmake.python_3x import StringIO, is_string, read_input, subprocess_output_to_string
+    from deepomatic.dmake.python_3x import StringIO, is_string, to_string, read_input, subprocess_output_to_string
 else:
-    from deepomatic.dmake.python_2x import StringIO, is_string, read_input, subprocess_output_to_string
+    from deepomatic.dmake.python_2x import StringIO, is_string, to_string, read_input, subprocess_output_to_string
 
 ###############################################################################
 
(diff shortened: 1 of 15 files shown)


deepomatic/dmake/deepobuild.py
```diff
@@ -923,10 +923,12 @@ def _get_service_(self, service):
...
     def _get_link_opts_(self, commands, service):
-        if common.options.dependencies and service.tests.has_value():
-            docker_links_names = service.tests.docker_links_names
-            if len(docker_links_names) > 0:
-                append_command(commands, 'read_sh', var = 'DOCKER_LINK_OPTS', shell = 'dmake_return_docker_links %s %s' % (self.app_name, ' '.join(docker_links_names)), fail_if_empty = True)
+        if common.options.dependencies:
+            needed_links = service.needed_links
+            if service.tests.has_value(): # deprecated
+                needed_links += service.tests.docker_links_names
+            if len(needed_links) > 0:
```
thomas-riccardi: maybe raise an error if both new and deprecated fields are used in parallel?

deepomatic/dmake/common.py
```diff
@@ -32,6 +34,29 @@ def __init__(self):
 
 ###############################################################################
 
+def yaml_ordered_load(stream, Loader=yaml.Loader, object_pairs_hook=OrderedDict):
```
thomas-riccardi: Wouldn't https://pypi.python.org/pypi/ruamel.yaml/ be better ?

PR #82, opened 2017-11-27
More pythonic structure: remove deepomatic directory

vdel:
Python package is now directly defined at the root of the repository.


thomas-riccardi:
what happened here?

PR #83, opened 2017-11-27
Nvidia docker fallback

vdel:
When nvidia-docker is not available, we fallback on docker and cross fingers.

### DIFF
--- a/deepomatic/dmake/deepobuild.py
+++ b/deepomatic/dmake/deepobuild.py
@@ -390,7 +390,8 @@ class SSHDeploySerializer(YAML2PipelineSerializer):
               ('export MID_DEPLOY_HOOKS="%s" && ' % config.mid_deploy_script) + \
               ('export POST_DEPLOY_HOOKS="%s" && ' % config.post_deploy_script) + \
               ('export READYNESS_PROBE="%s" && ' % common.escape_cmd(config.readiness_probe.get_cmd())) + \
-              ('export DOCKER_CMD="%s" && ' % ('nvidia-docker' if config.need_gpu else 'docker')) + \
+              ('export INSTALL_NVIDIA_DRIVERS="%s" && ' % ('1' if config.need_gpu else '')) + \
+              ('export DOCKER_RUN_CMD="%s" && ' % ('nvidia-docker' if config.need_gpu and common.has_nvidia_docker else 'docker')) + \
                'dmake_copy_template deploy/deploy_ssh/start_app.sh %s' % start_file
         common.run_shell_command(cmd)
 
@@ -946,12 +947,10 @@ class DMakeFile(DMakeFileSerializer):
 
     def generate_run(self, commands, service_name, docker_links, service_customization):
         service = self._get_service_(service_name)
-        if service.config is None or service.config.docker_image.start_script is None:
+        if not service.config.has_value() or service.config.docker_image.start_script is None:
             return
 
-        docker_run_prefix = ''
-        if service.config.need_gpu:
-            docker_run_prefix = 'DOCKER_CMD=nvidia-docker '
+        docker_run_prefix = "DOCKER_RUN_CMD='%s'" % common.get_docker_run_cmd(service.config.need_gpu)
 
         unique_service_name = service_name
         customized_env = {}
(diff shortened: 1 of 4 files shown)


deepomatic/dmake/deepobuild.py
```diff
@@ -390,7 +390,8 @@ def _serialize_(self, commands, app_name, config, image_name, env):
               ('export MID_DEPLOY_HOOKS="%s" && ' % config.mid_deploy_script) + \
               ('export POST_DEPLOY_HOOKS="%s" && ' % config.post_deploy_script) + \
               ('export READYNESS_PROBE="%s" && ' % common.escape_cmd(config.readiness_probe.get_cmd())) + \
-              ('export DOCKER_CMD="%s" && ' % ('nvidia-docker' if config.need_gpu else 'docker')) + \
+              ('export INSTALL_NVIDIA_DRIVERS="%s" && ' % ('1' if config.need_gpu else '')) + \
+              ('export DOCKER_RUN_CMD="%s" && ' % ('nvidia-docker' if config.need_gpu and common.has_nvidia_docker else 'docker')) + \
```
thomas-riccardi: Shouldn't the detection be done on the target ssh machine instead of the dmake one (for `common.has_nvidia_docker` on `SSHDeploySerializer`) ?

deepomatic/dmake/templates/deploy/deploy_ssh/start_app.sh
```diff
@@ -34,15 +34,15 @@ if [ -e $device_name ] && [ `df | grep $device_name | wc -l` == "0" ] && [ ! -e
 fi
 
 # Install NVIDIA drivers
-if [ "${DOCKER_CMD}" = "nvidia-docker" ]; then
+if [ ! -z "${INSTALL_NVIDIA_DRIVERS}" ]; then
```
thomas-riccardi: much better than only one env var !

deepomatic/dmake/templates/deploy/deploy_ssh/start_app.sh
```diff
@@ -34,15 +34,15 @@ if [ -e $device_name ] && [ `df | grep $device_name | wc -l` == "0" ] && [ ! -e
...
     if [ ! `which nvidia-smi` ]; then
         apt-get update
         apt-get install nvidia-375
     fi
     # Make sure modules are loaded
     modprobe nvidia
     modprobe nvidia_uvm
-    if [ ! `which nvidia-docker` ]; then
+    if [[ "${DOCKER_RUN_CMD}" =~ nvidia-docker.* ]]; then
```
thomas-riccardi: OK, I'll need to adapt #73

vdel: oops, I just merged #73, I will adapt in this PR

deepomatic/dmake/deepobuild.py
vdel: Hum, I don't think so: on the target SSH the assumption is that we do what it needs to run in the correct environment.
However, I do agree that it's the ssh script on the target that needs to decide of it's own DOCKER_RUN_CMD depending on the value of INSTALL_NVIDIA_DRIVERS

thomas-riccardi: that's what I meant: dmake-local `has_nvidia_docker` has no meaning in the remote ssh machine.

maingoh:
How is it going ? 
I have a crash in worker NN of this kind due to my GPU (this crash is weird though, I think alexis got the same on acacia)
```
F1218 18:33:40.247404     1 common.cpp:323] Check failed: error == cudaSuccess (30 vs. 0)  unknown error
*** Check failure stack trace: ***
    @     0x7f34a48b55cd  google::LogMessage::Fail()
    @     0x7f34a48b7433  google::LogMessage::SendToLog()
    @     0x7f34a48b515b  google::LogMessage::Flush()
    @     0x7f34a48b7e1e  google::LogMessageFatal::~LogMessageFatal()
    @     0x7f34a348d242  caffe::Caffe::SetDevice()
    @     0x7f34a6755756  setDevice()
    @     0x7f34a6755a36  deepomatic::CaffeNet::load()
    @     0x7f34a6759d03  deepomatic::CaffeNetLoader::load()
    @     0x7f34a67fa51b  deepomatic::NetFactory::create()
    @     0x7f34a67ea861  deepomatic::InferenceStep::InferenceStep()
    @     0x7f34a67c348c  WorkerNN::loadPipeline()
    @     0x7f34a67c137a  WorkerNN::WorkerNN()
    @           0x43d6c0  main
    @     0x7f34a208b830  __libc_start_main
    @           0x43bc99  _start
    @              (nil)  (unknown)
```

I have the same version of the drivers inside and outside the container. I want to bypass the GPU on my desktop machine to continue working with CPU. Even with `need_gpu: false` it uses my GPU inside the docker.
Also I would like to avoid commiting by mistake the `need_gpu: false`. 
An environnement variable would work

thomas-riccardi:
resolved in #140

PR #85, opened 2017-11-29
Update tutorial with newly deprecated field `docker_links_names`.

thomas-riccardi:


vdel:
You are right, leaving the deprecated `docker_link_name` is confusing in the tutorial.

I had left them on purpose because it also serves as tests. We should introduce separate tests.

Issue #87, opened 2017-11-29
Don't write dmake.yml file on applied migrations in Jenkins


thomas-riccardi:
When run by a bot, writing a new version of the dmake.yml file is useless: nobody will review and commit that.
Furthermore, it's best practice to *not* write into the worktree unless strictly needed.

PR #88, opened 2017-11-29
Test DMake with Python 2 and 3 in parallel.

thomas-riccardi:
Need to use different workspaces because dmake takes too man liberties on writing in the worktree (migrations (cf #87), but not only: `build` works in the local worktree mounted in a tmp container, should probably be removed too).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -30,17 +30,6 @@ node {
     env.PYTHONPATH = "${env.WORKSPACE}:${env.PYTHONPATH}"
     env.PATH = "${env.WORKSPACE}/deepomatic/dmake:${env.WORKSPACE}/deepomatic/dmake/utils:${env.PATH}"
 
-    // Clone repo to test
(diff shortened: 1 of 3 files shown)


thomas-riccardi:
Need to fix `Using the ‘stage’ step without a block argument is deprecated` jenkins warning before merging this PR, because with parallel build this warning becomes an error it seems.

thomas-riccardi:
We have deeper parallelism issues:

```
[python-3] + docker run --name dmake.parallel_python_2_3.10.dmake-tutorial.rabbitmq.0 -d -e RABBITMQ_DEFAULT_USER=user -e RABBITMQ_DEFAULT_PASS=password -e RABBITMQ_DEFAULT_VHOST=dev -i rabbitmq:3.6
...
[python-3] docker: Error response from daemon: Conflict. The container name "/dmake.parallel_python_2_3.10.dmake-tutorial.rabbitmq.0" is already in use by container "bd8a22d57833654f40022db1f01b2256cdd4cc5edeb65f2148ab49dd1400d0eb". You have to remove (or rename) that container to be able to reuse that name.
```

thomas-riccardi:
test

PR #89, opened 2017-11-29
Generated jenkinsfile stage block

thomas-riccardi:


thomas-riccardi:
Helps for #88

deepomatic/dmake/core.py
```diff
@@ -697,6 +711,9 @@ def make(root_dir, sub_dir, command, app, options):
                print(('ERROR in file %s:\n' % file) + str(e))
                sys.exit(1)
 
+        if len(commands) > 0:
```
vdel: You need to indent it inside the `for stage, commands in ordered_build_files:` loop.

By the way, I would actually clarify be doing:

```python
for stage, commands in ordered_build_files:
    if len(commands) == 0:
        continue
    append_command(all_commands, 'stage', name = stage, concurrency = 1 if stage == "Deploying" else None)
    for node, order in commands:
        pass # the rest of the code here
    append_command(all_commands, 'stage_end')
```

thomas-riccardi: indent is correct; added the `continue`

but maybe we could keep the stage with only one step: `echo Nothing to do` ?

PR #92, opened 2017-11-30
Proper handling of services with no deploy

vdel:
Currently, a service that:
- had no `deploy` configuration
- a default `config::docker_image::name` value 
led to dmake refusual to deploy.

This fixes it.

### DIFF
--- a/deepomatic/dmake/core.py
+++ b/deepomatic/dmake/core.py
@@ -665,7 +665,6 @@ def make(root_dir, sub_dir, command, app, options):
             append_command(all_commands, 'stage', name = stage, concurrency = 1 if stage == "Deploying" else None)
         for node, order in commands:
             command, service, service_customization = node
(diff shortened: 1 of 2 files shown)


deepomatic/dmake/core.py
```diff
@@ -674,29 +673,34 @@ def make(root_dir, sub_dir, command, app, options):
...
+                    dmake_file.generate_deploy(step_commands, service)
                 else:
                    raise Exception("Unkown command '%s'" % command)
             except DMakeException as e:
                print(('ERROR in file %s:\n' % file) + str(e))
                sys.exit(1)
 
+            if len(step_commands) > 0:
+                append_command(all_commands, 'echo', message = '- Running %s' % (display_command_node(node)))
+                all_commands += step_commands
```
thomas-riccardi: could add `else echo nothing to do` ?

PR #94, opened 2017-12-04
Execute tests using service docker image, not base docker image.

### LINKED ISSUE #34: Re-use deployment image to run tests
Currently the tests are `exec`'d on top of base_image, by first rebuilding with `build.commands` like for the deployment docker image build.
Furthermore, both test and deployment builds are done on the same source workspace, without clean, with no order guaranteed.

Instead, it would be far simpler to first build the deployment docker image, then use this deployment image to execute the tests.
It would bring:
- no useless rebuild
- guaranteed tests executed on the deployment image

Later:
- cleanup/explicit artifact handling: maybe remove tests executables from official image

thomas-riccardi:
Closes #34

Pros:
- test production/official service docker image instead of a parallel
  build (without install step)
- avoid duplicate build
- avoid writing build files in host workspace
- allows parallel build

Cons:
- need to ship tests programs (and its dependencies) in official image

TODO:
- [x] use service image to execute tests
- [x] execute tests all at once in one container so they share the environment like before
- [x] refactor code
- [x] access test results (using `docker cp`)
- [x] migrate tests users to not use `*` anymore (not supported anymore, to be able to do explicit `docker cp`)
- [x] fix vulcain `api-tmp` which has a `tests` but no `config`: it's not supported anymore, since `config.docker_image` describes how to build the service docker image 
- [x] everything still works

### DIFF
--- a/deepomatic/dmake/common.py
+++ b/deepomatic/dmake/common.py
@@ -147,7 +147,7 @@ def get_dmake_build_type():
 ###############################################################################
 
 def init(_command, _root_dir, _app, _options):
-    global root_dir, tmp_dir, config_dir, cache_dir, key_file
+    global root_dir, tmp_dir, config_dir, cache_dir, relative_cache_dir, key_file
     global branch, target, is_pr, pr_id, build_id, commit_id, force_full_deploy
     global repo_url, repo, use_pipeline, is_local, skip_tests, is_release_branch
     global build_description
@@ -160,7 +160,8 @@ def init(_command, _root_dir, _app, _options):
     config_dir = os.getenv('DMAKE_CONFIG_DIR', None)
     if config_dir is None:
         raise DMakeException("DMake seems to be badly configured: environment variable DMAKE_CONFIG_DIR is missing. Try to run %s again." % os.path.join(os.getenv('DMAKE_PATH', ""), 'install.sh'))
-    cache_dir = os.path.join(root_dir, '.dmake')
+    relative_cache_dir = '.dmake'
+    cache_dir = os.path.join(root_dir, relative_cache_dir)
     try:
         os.mkdir(cache_dir)
     except OSError:
--- a/deepomatic/dmake/deepobuild.py
+++ b/deepomatic/dmake/deepobuild.py
@@ -1,7 +1,7 @@
 import os
 import copy
 import json
-import random
+import uuid
 import importlib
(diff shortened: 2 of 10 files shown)


deepomatic/dmake/deepobuild.py
```diff
@@ -1194,23 +1196,41 @@ def _generate_test_docker_cmd_(self, commands, service, docker_links):
...
+            docker_opts += " " + (" ".join(opts))
+
+        env = self.env.get_replaced_variables(docker_links=docker_links, needed_links=service.needed_links)
+        image_name = service.config.docker_image.get_image_name(env = env)
+        docker_opts += " -i %s" % image_name
+
+        docker_cmd = 'dmake_run_docker_test %s "" %s ' % (service_name, docker_opts)
+
+        if service.config.has_value() and service.config.need_gpu:
+            docker_cmd = 'DOCKER_CMD=nvidia-docker ' + docker_cmd
```
vdel: Didn't we said it was docker everywhere with Nvidia runtime ?

deepomatic/dmake/common.py
```diff
@@ -160,7 +160,8 @@ def init(_command, _root_dir, _app, _options):
     config_dir = os.getenv('DMAKE_CONFIG_DIR', None)
     if config_dir is None:
         raise DMakeException("DMake seems to be badly configured: environment variable DMAKE_CONFIG_DIR is missing. Try to run %s again." % os.path.join(os.getenv('DMAKE_PATH', ""), 'install.sh'))
-    cache_dir = os.path.join(root_dir, '.dmake')
+    relative_cache_dir = '.dmake'
```
vdel: Was it really needed to add a new global variable ?

deepomatic/dmake/deepobuild.py
thomas-riccardi: Fixed; I started working on this before the revert, and copied code, and didn't catch it upon rebase. (checking this was later on my todo list however)

deepomatic/dmake/common.py
thomas-riccardi: Not great indeed but alternatives were worse:
- hardcoding `.dmake` everywhere: may be the best alternative, but then we'd just remove `common.cache_dir` also, shouldn't we?
- keep only the `common.relative_cache_dir` and remove `common.cache_dir`: recompute it when needed?
- compute relative path from full path: bad idea

what would be your solution?

thomas-riccardi:
Ready to merge; not yet to deploy on Jenkins.

PR #103, opened 2017-12-14
Allow to run dmake without internet connection

vdel:
This works only if base images are provided locally.

### DIFF
--- a/deepomatic/dmake/deepobuild.py
+++ b/deepomatic/dmake/deepobuild.py
@@ -3,6 +3,7 @@ import copy
 import json
 import random
 import importlib
+import requests.exceptions
 from deepomatic.dmake.serializer import ValidationError, FieldSerializer, YAML2PipelineSerializer
 import deepomatic.dmake.common as common
 from deepomatic.dmake.common import DMakeException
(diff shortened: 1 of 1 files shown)


deepomatic/dmake/deepobuild.py
```diff
@@ -186,7 +187,19 @@ def _serialize_(self, commands, path_dir):
...
-            root_image_digest = docker_registry.get_image_digest(self.root_image)
+            try:
+                root_image_digest = docker_registry.get_image_digest(self.root_image)
+            except requests.exceptions.ConnectionError as e:
+                if not common.is_local:
+                    raise e
+                common.logger.info('Could not reach registry, you seem to be offline.')
+                common.logger.info('Trying to find {} locally'.format(self.root_image))
+                try:
+                    response = common.run_shell_command('docker inspect {}'.format(self.root_image))
```
thomas-riccardi: could use newer cli: `docker image inspect`

vdel:
Rebuild is usually not possible: you need to download packages

Le jeu. 14 déc. 2017 à 14:10, Thomas Riccardi <notifications@github.com> a
écrit :

thomas-riccardi:
Indeed.
OK so if base image sources have changed, dmake_build_base_docker will fail
during hub search of base image build cache, and the best we can do then is
to print a better error explaining we can't get hub version, and we don't
want to try to build it locally as we are probably offline.

Also, the log info could be improved: explaining that the root image cannot
be checked if it is up-to-date because of probably offline; thus using the
local potentially outdated root_image.

Maybe we should instead add an explicit offline mode (DMAKE_OFFLINE), that
doesn't even try to use internet, and print warnings everywhere it skips
internet calls.

vdel:
> print a better error explaining we can't get hub version, and we don't
want to try to build it locally as we are probably offline.

OK

> Also, the log info could be improved: explaining that the root image cannot
be checked if it is up-to-date because of probably offline; thus using the
local potentially outdated root_image.

OK

> Maybe we should instead add an explicit offline mode (DMAKE_OFFLINE), that
doesn't even try to use internet

I think we should prefer simplicity over control. We are power users, others in the team do not know what dmake is doing, we should not bother them with configurations to edit.

> and print warnings everywhere it skips internet calls.

I think that in this case, either you have all the base images locally (which happens most of the time), or you are screwed.

thomas-riccardi:
> > print a better error explaining we can't get hub version, and we don't
> want to try to build it locally as we are probably offline.
> 
> So, you're speaking about this line ? https://github.com/Deepomatic/dmake/blob/ignore_missing_connection/deepomatic/dmake/utils/dmake_build_base_docker#L63
> How do you know we cannot reach the hub ? Both error codes in case of non-existing image (e.g. for a new base image) and un-reachable hub are the same (i.e. error code 1). Error messages are not the same however:
> - in case of non existing image: `Error response from daemon: repository bloublou123 not found: does not exist or no pull access`
> - in case of error on hub: `Error response from daemon: Get https://registry-1.docker.io/v2/: dial tcp: lookup registry-1.docker.io on X.X.X.X: no such host`
> 
> But I would not rely on error messages to test that. 
> We could ping the hub, but what about images from a private repository ? We need to extract the hub address from the image.
> 
> I find it's getting complicated for just a warning message and would maybe require a separate issue ?

OK, opened #109

PR #104, opened 2017-12-14
Evaluate needed service env in context

vdel:
Needed service env used to be evaluated irrespective of variables exported by links, which led to wrong variable replacement.

### DIFF
--- a/deepomatic/dmake/deepobuild.py
+++ b/deepomatic/dmake/deepobuild.py
@@ -792,12 +792,6 @@ class NeededServiceSerializer(YAML2PipelineSerializer):
         super(NeededServiceSerializer, self).__init__(**kwargs)
         self._specialized = True
 
-    # def __str__(self):
-    #     s = "%s" % (self.service_name)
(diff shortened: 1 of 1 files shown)


deepomatic/dmake/deepobuild.py
```diff
@@ -792,11 +792,11 @@ def __init__(self, **kwargs):
...
-    def __str__(self):
-        s = "%s" % (self.service_name)
-        if self._specialized:
-            s += " -- env: %s" % (self.env)
-        return s
+    # def __str__(self):
+    #     s = "%s" % (self.service_name)
+    #     if self._specialized:
+    #         s += " -- env: %s" % (self.env)
+    #     return s
```
thomas-riccardi: why ?

thomas-riccardi: I don't remember why I added this, maybe it was for pretty print during execution of steps?
Probably not needed anymore, replaced by `display_command_node()` in `core.py`.

Just remove these lines then.

PR #105, opened 2017-12-14
Avoid deploying all pods at the same time

vdel:
Improve stability of the cluster

### DIFF
--- a/deepomatic/dmake/utils/dmake_deploy_k8s_cd
+++ b/deepomatic/dmake/utils/dmake_deploy_k8s_cd
@@ -128,19 +128,18 @@ for i in ret.items:
     beta_v1_client.patch_namespaced_deployment(name, namespace, json_patch)
     deployments.append(name)
 
-# # Actually not a good solution: we would need to remove those entries if this file succeed
-# # Removing is not so easy as there might some concurreny writes in this file
-# # As it is unlikely to fail here anyway, for now we do not roll back deployment in case of failure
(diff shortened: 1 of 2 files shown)


thomas-riccardi:
Hugo proposes to use gevent with a controlled pool size to avoid waves of start & wait in chunks and have instead a constant number of in flight jobs:
```
from gevent.monkey import patch_all
patch_all()
from gevent.pool import Pool
import time
import random


def deploy_worker_nn(i):
    print "deploying worker n %s" % i
    sleep_time = random.uniform(0.5, 2.0)
    time.sleep(sleep_time)  # worker takes between 0.5s and 2.0 to deploy


pool = Pool(10)

for i in range(10000):
    pool.spawn(deploy_worker_nn, i)  # will block if pool is full


pool.join()
```
caveats: 
- monkey patching need to be done early in the python process life
- gevent + kubernetes client to test

maingoh:
According to the requirements, it's just a REST client that uses requests so I am confident it will work with gevent.

vdel:
That's completely overkill no ?
I would use gevent very carefully and the least possible

maingoh:
A threadpool would also work. I didn't know exactly what it was for but thomas asked for something in python that control the maximum of concurrent tasks

thomas-riccardi:
it is indeed probably overkill; I wanted to avoid: {update N; wait N;} loop
and have alwas N updating, it's faster; but it's clearly not a priority.

vdel:
Up on this: it's actually an important fix that avoids apocalypse when deploying worker-nn

PR #110, opened 2017-12-29
Execute tests using service docker image, not base docker image.

### LINKED ISSUE #34: Re-use deployment image to run tests

thomas-riccardi:
Closes #34

Pros:
- test production/official service docker image instead of a parallel
  build (without install step)
- avoid duplicate build
- avoid writing build files in host workspace
- allows parallel build

Cons:
- need to ship tests programs (and its dependencies) in official image

TODO:
- [x] use service image to execute tests
- [x] execute tests all at once in one container so they share the environment like before
- [x] refactor code
- [x] access test results (using `docker cp`)
- [x] migrate tests users to not use `*` anymore (not supported anymore, to be able to do explicit `docker cp`)
- [x] fix vulcain `api-tmp` which has a `tests` but no `config`: it's not supported anymore, since `config.docker_image` describes how to build the service docker image 
- [x] everything still works

### DIFF
--- a/deepomatic/dmake/common.py
+++ b/deepomatic/dmake/common.py
@@ -147,7 +147,7 @@ def get_dmake_build_type():
 ###############################################################################
 
 def init(_command, _root_dir, _app, _options):
-    global root_dir, tmp_dir, config_dir, cache_dir, key_file
+    global root_dir, tmp_dir, config_dir, cache_dir, relative_cache_dir, key_file
     global branch, target, is_pr, pr_id, build_id, commit_id, force_full_deploy
     global repo_url, repo, use_pipeline, is_local, skip_tests, is_release_branch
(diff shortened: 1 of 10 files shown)


thomas-riccardi:
New version of #94, with more code, and **please merge, not squash**, and rebased on top of newer master.

PR #116, opened 2018-01-12
Install readme update

thomas-riccardi:


install.sh
```diff
@@ -147,8 +147,9 @@ if [ -z "`which dmake`" ]; then
             echo "Patched ${SHRC} to source ${CONFIG_FILE}."
         fi
     done
-    echo "You should be good to go after restarting a shell session !"
+    echo "Install python dependencies with the following command:"
+    echo "pip install --user -r requirements.txt"
```
vdel: Actually, when we add a new python dependancy we should re-run ```pip```
What about running it automatically, outside of this `if` ?

thomas-riccardi: It's just an echo for no, I don't want to default install without an isolated environment (virtualenv, pipenv..).
Added a reminder on non-first install.

PR #134, opened 2018-02-13
Add `volumes` at dmake.yml root: declare shared volumes by name

### LINKED ISSUE #135: dmake shared volumes between services
We need to support explicit (empty) shared volumes between services (both app services and docker_links services).

Current workaround are hacky and create tons of issues.

thomas-riccardi:
Closes #135 

They can be used during the dmake session in any service (app service
or docker_link service).

They are deleted at the end of the dmake session.

Permissions in the create volumes are set to
`${DMAKE_UID}:${DMAKE_UID}` so app services can write; it may require
manual handling for permissions with `docker_link` services.

Usage:
=====

- `services[].config.volumes` now accept named shared volumes

In addition to object {container_volume, host_volume}, it now accepts
a string value: `<volume_name>:<container_path>`

The `<volume_name>` is the name of a shared volume declared in root
`volumes`.

- `docker_link[].volumes` now also accepts named shared volumes

In addition to `<host_path>:<container_path>` (which is still only
applied for shell, because we don't want persistence during tests), it
now accepts: `<volume_name>:<container_path>` (always applied (shell,
run, test)).

The `<volume_name>` is the name of a shared volume declared in root
`volumes`.

Merged implicit and explicit old volumes types for `docker_links` and
`services`: `VolumeMountSerializer`
Now both accept both object and string, but keep their specific
behavior for now (no breaking change):
- `docker_link`:
  - ignored when not in shell (we don't want persistence during tests:
  use named volumes if need to share volumes between services)
  - relative host paths are relative to dmake file
- `service`:
  - when used locally (not in ssh deploy), force host path to be
  relative to root_repo/.dmake/volumes, not sure why

Examples:
```
    volumes:
      - ./foo1:/foo1
      - /foo2:/foo2
      - container_volume: ./foo3
        host_volume: /foo3
      - container_volume: /foo4
        host_volume: /foo4
```
`foo1` and `foo3` forms are equivalent; `foo2` and `foo4` are
equivalent.

New alternative type for volumes: named shared volumes: `SharedVolumeMountSerializer`
Examples:
```
    volumes:
      - foo5:/foo5
      - source: foo6
        target: /foo6
```
`foo5` and `foo6` forms are equivalent.

See also `tutorial/` and deploy/test-shared-volumes.sh

Details:
=======

Create shared volumes on demand using core.py dependency resolution,
by creating 2 objects, and links between them:
- `SharedVolumeSerializer`: the shared volume declaration in root
  `volumes
- `SharedVolumeMountSerializer` the shared volume usage on
  `docker_link` and `service`

Services and docker_links depend on creating the shared volumes for
run, shell, test commands.
Added new command: `shared_volume`.

### DIFF
--- a/deepomatic/dmake/common.py
+++ b/deepomatic/dmake/common.py
@@ -4,6 +4,7 @@ import logging
 import subprocess
 import re
 from ruamel.yaml import YAML
+import uuid
 
 # Set logger
 logger = logging.getLogger("deepomatic.dmake")
@@ -30,6 +31,11 @@ class DMakeException(Exception):
 class NotGitRepositoryException(DMakeException):
     def __init__(self):
         super(NotGitRepositoryException, self).__init__('Not a GIT repository')
+
+class SharedVolumeNotFoundException(DMakeException):
+    def __init__(self, name):
+        super(SharedVolumeNotFoundException, self).__init__("Unknown volume named '%s'" % (name))
+        self.name = name
 
 ###############################################################################
 
@@ -168,9 +174,12 @@ def init(_command, _root_dir, _app, _options):
     global build_description
     global command, options, uname
     global do_pull_config_dir
+    global session_id
     root_dir = os.path.join(_root_dir, '')
     command = _command
     options = _options
+
+    session_id = uuid.uuid4()
 
     config_dir = os.getenv('DMAKE_CONFIG_DIR', None)
     if config_dir is None:
--- a/deepomatic/dmake/core.py
+++ b/deepomatic/dmake/core.py
@@ -2,7 +2,7 @@ import os, sys
 import uuid
 
 import deepomatic.dmake.common as common
-from   deepomatic.dmake.common import DMakeException
+from   deepomatic.dmake.common import DMakeException, SharedVolumeNotFoundException
 from   deepomatic.dmake.deepobuild import DMakeFile, append_command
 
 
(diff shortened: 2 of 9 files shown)


thomas-riccardi:
So, manual builds of https://github.com/Deepomatic/vesta/pull/837 with this dmake PR is hard because vesta tests are unstables: sometimes it's training-worker digits tests that fail, some other times celery, and dmake is tested both on python 2 and python 3: it's statistically hard to get both working in one job. But : on the cumulated jobs every vesta test passed both on dmake python 2 and 3.

We should be good to go.

thomas-riccardi:
We still have permissions issues on the volume: except for `dmake shell` everything is run as root in container: OK; but for `dmake shell` we still have the `DMAKE_UID` feature that allows to generate files from the container into the host with host user permissions: this feature breaks permissions across containers in shared volumes because only the shell service is non-root in container...

Either we drop `DMAKE_UID`, or we `chmod 777` (and `umask`) the shared volumes, or we try uid-remapping so that container root is equal to host user...

thomas-riccardi:
For the uid issue https://boxboat.com/2017/07/25/fixuid-change-docker-container-uid-gid/ could help but requires startup script specific to shell, and extra program installed at build time, and forcing user to non root in container (reverse that what we wanted: root in container, uid standard in host).

uid-remap would probably work (not sure if it can be used to map *all* container uids to just one host uid: the current user id), but would be limited to one user per service as the uid-remap is global to the docker daemon, not per docker run.

thomas-riccardi:
more user-remap test:

With:
```
$ cat /etc/subuid
thomas:1000:1
```
it breaks everything: docker run fails:
```
docker: failed to register layer: Error processing tar file(exit status 1): Container ID 104 cannot be mapped to a host ID.
See 'docker run --help'.
```
The goal was to try to force all container ids to just one uid on host.
when allowing 1:1 mapping, it works, but container uid 1 is mapped to host uid 1001: not great at all...

thomas-riccardi:
Went for `chmod 777` when using `dmake shell`.
Maybe we should always do that... it's currently not needed as we run everything as root; but later we may want to change users at build time; in which case the chmod 777 would help even for non shell.

PR #138, opened 2018-02-19
Add variables to k8s deploy template manifest

thomas-riccardi:
Add optional template variables for kubernetes deployment manifest

Usage:
- in `dmake.yml`:
```
env:
  branches:
    default:
      variables:
        K8S_DEPLOY_TLS_SECRET_NAME: dev-tls-secret
    master:
      variables:
        K8S_DEPLOY_TLS_SECRET_NAME: prod-tls-secret
...
services:
  - deploy:
      stages:
        - kubernetes:
            manifest:
              template: deploy/kubernetes-deploy.yaml
              variables:
                TLS_SECRET_NAME: ${K8S_DEPLOY_TLS_SECRET_NAME}
```
- in `deploy/kubernetes-deploy.yaml`:
```

apiVersion: extensions/v1beta1
kind: Ingress
...
spec:
  tls:
    - secretName: ${TLS_SECRET_NAME}
  backend:
    serviceName: web
    servicePort: 80
```

Still support the simple version where
`services[].deploy.stages[].kubernetes.manifest` is directly a file
path to the manifest.

Also add optional strict mode for `common.eval_str_in_env`: raise error when using undefined variable: using it in `manifest.variables` to avoid typo errors that convert to empty strings, which is valid in k8s manifest: error would happen too late in the cluster.

### DIFF
--- a/deepomatic/dmake/common.py
+++ b/deepomatic/dmake/common.py
@@ -90,10 +90,13 @@ def escape_cmd(cmd):
 def wrap_cmd(cmd):
     return '"%s"' % cmd.replace('"', '\\"')
 
-def eval_str_in_env(cmd, env=None):
+def eval_str_in_env(value, env=None, strict=False):
     if env is None:
         env = {}
-    cmd = 'echo %s' % wrap_cmd(cmd)
+    cmd = ''
+    if strict:
+        cmd += 'set -euo pipefail; '
+    cmd += 'echo %s' % wrap_cmd(value)
     return run_shell_command(cmd, additional_env=env).strip()
 
 # Docker has some trouble mounting volumes with trailing '/'.
--- a/deepomatic/dmake/deepobuild.py
+++ b/deepomatic/dmake/deepobuild.py
@@ -515,10 +515,31 @@ class KuberentesConfigMapSerializer(YAML2PipelineSerializer):
     name = FieldSerializer("string", example = "nginx", help_text = "Kubernetes ConfigMap name")
     from_files = FieldSerializer("array", child = KuberentesConfigMapFromFileSerializer(), default = [], help_text = "Kubernetes ConfigMap from files")
 
+class KubernetesManifestSerializer(YAML2PipelineSerializer):
+    template  = FieldSerializer("file", example = "path/to/kubernetes-manifest.yaml", help_text = "Kubernetes manifest file (template) defining all the resources needed to deploy the service")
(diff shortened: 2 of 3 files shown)


deepomatic/dmake/common.py
```diff
@@ -90,10 +90,13 @@ def escape_cmd(cmd):
...
     return '"%s"' % cmd.replace('"', '\\"')
 
-def eval_str_in_env(cmd, env=None):
+def eval_str_in_env(value, env=None, strict=False):
     if env is None:
         env = {}
-    cmd = 'echo %s' % wrap_cmd(cmd)
+    cmd = ''
+    if strict:
+        cmd += 'set -euo pipefail; '
```
vdel: does it work in all shells ?

deepomatic/dmake/deepobuild.py
```diff
@@ -515,10 +515,31 @@ class KuberentesConfigMapSerializer(YAML2PipelineSerializer):
...
+class KubernetesManifestSerializer(YAML2PipelineSerializer):
+    template  = FieldSerializer("file", example = "path/to/kubernetes-manifest.yaml", help_text = "Kubernetes manifest file (template) defining all the resources needed to deploy the service")
+    variables = FieldSerializer('dict', child="string", default={}, help_text="Defines variables used in the kubernetes manifest template", example={'TLS_SECRET_NAME': '${K8S_DEPLOY_TLS_SECRET_NAME}'})
+
+    def _validate_(self, file, needed_migrations, data, field_name):
+        # also accept simple variant where data is a string: file path to the kubernetes manifest template
+        if common.is_string(data):
+            data = {'template': data}
+
+        result = super(KubernetesManifestSerializer, self)._validate_(file, needed_migrations=needed_migrations, data=data, field_name=field_name)
```
vdel: `return` directly ?

deepomatic/dmake/common.py
thomas-riccardi: we explicitly use `bash` which supports `-e` and `-o pipefile` at least since bash 3, `-u` at least since bash 4.2.

deepomatic/dmake/deepobuild.py
thomas-riccardi: I use this new pattern in all new `_validate_()`, it minimizes the diff when adding extra validation.
It's not ideal because older `_validate_()` return `self`, but at least it start normalizing onto something.

PR #141, opened 2018-02-20
Support `need_gpu` on all services, for all local executions.

### LINKED ISSUE #140: Support GPU locally with nvidia docker v2
Services (app and docker link) should be able to use a GPU using nvidia docker v2.

Currently `need_gpu` on a service is only used in the ssh deploy; and was tentatively used locally to y #72 which was soon reverted-back (in #97 ).
Previous attempt imposed nvidia docker v2 installed everywhere, and set as default runtime engine (which created its own issues) (see #76 for example).

New proposed solution:
- just require nvidia docker v2 runtime installed, not by default: use `docker run --runtime=nvidia` (and https://github.com/nvidia/nvidia-container-runtime#nvidia_visible_devices)
- be able to disable all this if user wants to (for example no GPU is available locally), without patching the `dmake.yml` files

thomas-riccardi:
Closes #140 

- added `need_gpu` on `docker_links[]` (in addition to previous
  `services[].config`)
- use `need_gpu` on local executions: run, run_link, test, shell
  (in addition to previous ssh deploy)

Locally we now use nvidia docker v2 with:
`docker run --runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=all`

This allows using GPU on any docker image (not just `FROM
nvidia/cuda`).

If local machine doesn't support nvidia docker v2, then GPU support
can be disabled with `export DMAKE_NO_GPU=1`: the services will be run
without GPU, it could break things: it's the user's responsibility.

### DIFF
--- a/deepomatic/dmake/deepobuild.py
+++ b/deepomatic/dmake/deepobuild.py
@@ -72,6 +72,18 @@ def generate_env_file(tmp_dir, env):
             f.write('%s=%s\n' % (key, value))
     return file
 
+###############################################################################
+
+def get_docker_run_gpu_cmd_prefix(need_gpu, service_type, service_name):
+    prefix = ''
+    if need_gpu:
+        if common.no_gpu:
+            common.logger.info("GPU needed by %s '%s' but DMAKE_NO_GPU set: trying without GPU." % (service_type, service_name))
+            pass
+        else:
+            prefix = 'DMAKE_DOCKER_RUN_WITH_GPU=all '
+    return prefix
+
 # ###############################################################################
 
 class EnvBranchSerializer(YAML2PipelineSerializer):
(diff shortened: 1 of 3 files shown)


deepomatic/dmake/deepobuild.py
```diff
@@ -969,6 +981,16 @@ def create_variant(self, variant):
 
         return service
 
+    def get_docker_run_gpu_cmd_prefix(self):
```
vdel: factorize this and the function above in a common function ?

thomas-riccardi: yes, but where ? Probably not in `common.py`, and inheritance won't work: `need_gpu` is not on the same level on both services..
Also, only the string `DMAKE_DOCKER_RUN_WITH_GPU=all '` is effectively factorizable, all the rest is subtly different because both types of services are subtly different.

thomas-riccardi: done

PR #143, opened 2018-02-21
Add a Jenkins lock on 'GPU' when testing services that need one

### LINKED ISSUE #142: Jenkins: add a resource lock for GPU on executions that need it
Currently GPU tasks are all launched in parallel, and we often reach VRAM limits: tests fail randomly with hard to understand errors.

With a jenkins resource lock on GPU, it would allow us to serialize the GPU usages: no more VRAM full.

The goal would be to lock the GPU resource on a per dmake execution basis, not per service (we want to share the one GPU per machine on all the services that need it in one dmake execution).

Use https://wiki.jenkins.io/display/JENKINS/Lockable+Resources+Plugin

thomas-riccardi:
Closes #142 

In practice it wraps the whole `Testing App` stage with a
`lock('GPU'){...}` block if any used service (app or docker_link)
requested a gpu via `need_gpu`.

Limitations:
- `dmake shell` is not supported: it is currently split into the
`Deploy` stage instead of `Testing App`, so it's harder to include it
in the lock, and `dmake shell` is never used with the Jenkins pipeline
mode anyway.
- locking is not minimal: we lock the whole testing stage instead of
just when services needing a GPU are up; but it doesn't change much:
we currently have only optimal service start, not service stop, so we
would not gain much. Also, services dependencies are recursive, so we
often need the GPU the whole time anyway.

### DIFF
--- a/deepomatic/dmake/common.py
+++ b/deepomatic/dmake/common.py
@@ -164,7 +164,7 @@ def init(_command, _root_dir, _app, _options):
     global root_dir, tmp_dir, config_dir, cache_dir, relative_cache_dir, key_file
     global branch, target, is_pr, pr_id, build_id, commit_id, force_full_deploy
     global repo_url, repo, use_pipeline, is_local, skip_tests, is_release_branch
-    global no_gpu
+    global no_gpu, need_gpu
     global build_description
     global command, options, uname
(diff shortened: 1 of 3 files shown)


vdel:
Nice !

PR #148, opened 2018-03-01
K8s deploy prune

thomas-riccardi:
Add pruning to kubernetes deployment

Previously when removing a resource it leaked in the cluster, because
`kubectl apply` cannot know for sure what to remove.

The way to do it is to explicitly ask for pruning, and limit scope to
some resources via label selectors.
For it to work, we need to
- set a specific label to all resources (top level only because
  cascading handles that (e.g. Deploy -> RC -> Pod)), that
  scopes to the current dmake deployment: dmake app service_name:
  `dmake.deepomatic.com/service=<app>-<service_name>`
  (later it could be changed to deployment, when we add the concept of
  deployment independent of dmake service; the transition may be hard
  though.)
- and select on that when pruning (and when applying: both
  are done at the same time and both are impacted by the label
  selection).

The easiest way is to load the generate yamls, patch them, and
re-dump them, because we have multiple resources:
- env configmap: dmake generates it: we add proper labels directly
- user configmaps from files: dmake generates it but via
  `kubectl create configmap`, which doesn't support adding labels
  yet (see kubernetes/kubernetes#60295):
  we patch the yaml file.
- user manifest: fully under user control: we patch the yaml file.

For easier rollback at kubernetes level, we want to keep old env
configmaps. For that we:
- add a second label: `dmake.deepomatic.com/prune=no-pruning`
  (could add other values later if needed)
  for the dmake generated env configmap, to distinguish them from
  other resources in the same apply scope (dmake service)
- `kubectl apply` in two steps:
  - first for non-pruning manifests (the dmake generated env
    configmap), without `--prune`
  - second for all other manifests, with `--prune` (user
    manifests + user configmaps from files)
(We are forced to do two `kubectl apply` because the label selectors
apply both to the apply and prune parts.)

Bonus:
 Now wait all deployments rollouts instead of just the ~main one

Previously dmake imposed naming the "main" deployment ==
${SERVICE_NAME}, and waited only this deployment rollout.

Now we support multiple deployments: we discover them all with
the new injected label `dmake.deepomatic.com/service=${SERVICE_NAME}`,
then call `kubectl rollout status` on each (successively because
kubectl doesn't support multi resources for now).

### DIFF
--- a/deepomatic/dmake/common.py
+++ b/deepomatic/dmake/common.py
@@ -39,26 +39,27 @@ class SharedVolumeNotFoundException(DMakeException):
 
 ###############################################################################
 
-def yaml_ordered_load(stream):
+def yaml_ordered_load(stream, all=False):
     try:
         yaml = YAML(pure=True)
-        data = yaml.load(stream)
+        data = list(yaml.load_all(stream)) if all else yaml.load(stream)
         return data
     except Exception as e:
         raise DMakeException(str(e))
 
-def yaml_ordered_dump(data, stream=None, default_flow_style=False):
-    yaml=YAML()
-    yaml.default_flow_style = default_flow_style
-    yaml.width = 4096
-    yaml.indent(mapping=2, sequence=4, offset=2)
-    string_value = StringIO()
-    yaml.dump(data, string_value)
-    string_value = string_value.getvalue()
-    if stream:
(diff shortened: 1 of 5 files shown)


thomas-riccardi:
*DONT* merge, need to merge #146 and rebase before.

PR #155, opened 2018-03-07
Fix dmake for python3

vdel:


vdel:
@thomas-riccardi : please merge when you are ready

PR #156, opened 2018-03-07
Use `dmake` as the root module

vdel:
Why ?

The `deepomatic` module is conflicting with the `deepomatic` of the client.
As a result, we cannot use it to test the client.

So we need to change either the client or dmake. I think it makes more sense to change dmake, thus leaving us the possibility to install dmake with `pip install dmake` and let our client install our SDK with `pip install deepomatic`.

People will need to adapt their `.dmake/config.sh` (by removing both `/deepomatic` from `export PATH=...`) to let dmake work.


install.sh
```diff
@@ -132,7 +132,7 @@ echo "export DMAKE_CONFIG_DIR=${DMAKE_CONFIG_DIR}" >> ${CONFIG_FILE}
 echo "export DMAKE_PULL_CONFIG_DIR=${DMAKE_PULL_CONFIG_DIR}" >> ${CONFIG_FILE}
 echo "export DMAKE_SSH_KEY=${DMAKE_SSH_KEY}" >> ${CONFIG_FILE}
 echo "export PYTHONPATH=\$PYTHONPATH:${DMAKE_PATH}" >> ${CONFIG_FILE}
-echo "export PATH=\$PATH:${DMAKE_PATH}/deepomatic/dmake/:${DMAKE_PATH}/deepomatic/dmake/utils" >> ${CONFIG_FILE}
+echo "export PATH=\$PATH:${DMAKE_PATH}/dmake/:${DMAKE_PATH}/dmake/utils" >> ${CONFIG_FILE}
```
thomas-riccardi: This breaking change could be avoided with symlinks: 
```
ln -s  {,deepomatic/}dmake/dmake
ln -s {,deepomatic/}dmake/utils
```
I mean, keep this line, but add retro-compatibility symlinks (it's OK as long as `deepomatic/` is *not* a python module anymore, isn't it?)

vdel: done

thomas-riccardi:
@vdel I pushed an additional commit on your branch, not sure if it's really needed though.

PR #162, opened 2018-03-09
Removing dependancy on base when no base image is provided

vdel:
Base images used to not be needed for simple use cases. This fixes a crash due to a bad dependency.

### DIFF
--- a/dmake/core.py
+++ b/dmake/core.py
@@ -162,6 +162,11 @@ def activate_service_shared_volumes(loaded_files, service_providers, service):
 ###############################################################################
 
 def activate_base(base_variant):
+    if base_variant is None:
+        # base_variant is None when no base image is specified,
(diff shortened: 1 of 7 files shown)


dmake/core.py
```diff
@@ -204,7 +204,10 @@ def activate_service(loaded_files, service_providers, service_dependencies, comm
             if with_dependencies and needs is not None:
                 children += activate_needed_services(loaded_files, service_providers, service_dependencies, needs)
                 children += activate_link(loaded_files, service_providers, service_dependencies, service)
-            children += activate_base(base_variant)
+            if base_variant is not None:
```
thomas-riccardi: should be moved in `activate_base`, notably there is another call to `activate_base(base_variant)` for command `build_docker`

Issue #77, opened 2017-11-21
deprecate `python_requirements`


thomas-riccardi:
`python_requirements` is a python specific helper for the base image, it should be removed: just copy the requirements.txt and install and run pip yourself.

PR #163, opened 2018-03-09
Set python3 as a low-priority alternative when using python3_requirements

vdel:
When using `python3_requirements` python3 is installed but calling `python` in the base image would not work as python2 is not installed and python3 is not declared as a python alternative.

We fix it by declaring python3 as a low-priority alternative.


thomas-riccardi:
closed becase `python3_requirements` should be deprecated: #77

Issue #161, opened 2018-03-08
Properly check dmake with python3: use python3 in dmake/utils too


thomas-riccardi:
We test dmake with python2 and python3 (see the `Jenkinsfile`) using virtualenvs; but in practice it only executes the main `dmake` program using python3.
There are other python scripts in dmake: (some of) the dmake utils (for example `dmake_replace_vars`).

We should:
- [ ] make the python dmake utils work on both python 2 and 3 (not sure about the shebang..)
- [ ] change the `Jenkinsfile` to run everything with python3 on the python3 test (probably use different docker node agents with a Dockerfile for each, instead of virtualenvs on the jenkins node directly, would even possibly be faster with the docker build cache for the dmake python requirements.txt installation).

thomas-riccardi:
The Dockerfile part becomes urgent: some python dependency required compilation, and we didn't have python3-dev installed in jenkins..

thomas-riccardi:
Following #434 we could support only python3.

thomas-riccardi:
Closing in favor of #434 and #210

PR #173, opened 2018-04-03
Fix python3 tests

vdel:


vdel:
I won't have the time to follow up on that as it is not ready, thus issue #161 still exists

PR #179, opened 2018-04-17
Jenkins Parameters for custom builds, evaluated for build args and labels

### LINKED ISSUE #178: Support manual custom builds with jenkins builds parameters and environment variables
We need to manually trigger custom builds with specific config, mainly for on-premises docker images.

The plan is to define environment variables to be used as build args through variable substitution by dmake.

thomas-riccardi:
Closes #178

* service docker_image.build: variable substitution for labels in addition to args

Useful to pass custom build args via environment variables.

Also applied "strict" mode on both `args` and `labels`: they are new
options, we should always use "strict" for new options.
"strict" mode raises an error if an undefined variable is used, or if
an error happens during evaluation.

If you want to allow an undefined variable, then use `${MY_VAR:-}`; it
will evaluate to empty string (we have no way to unset...).

Evaluation is done at serialization time: only if the service is
really built, so it should not be too cumbersome to use.

* Add CUSTOM_BUILD_* parameters in Jenkins builds

Used for manual builds of on-premises images with special build args
and labels.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -11,7 +11,13 @@ properties([
                description: 'Application to test. You can also specify a service name if there is no ambiguity. Use * to force the test of all applications.'),
         booleanParam(name: 'DMAKE_SKIP_TESTS',
                      defaultValue: false,
-                     description: 'Skip tests if checked')
+                     description: 'Skip tests if checked'),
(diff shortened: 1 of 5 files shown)


thomas-riccardi:
Deployed on jenkins for several days, merging.

PR #188, opened 2018-04-27
--command option for dmake shell

### LINKED ISSUE #186: Launch a command inside a dmake shell
I would like to do something like this : 
```
dmake shell myservice -c "python /app/script.py"
```
Where the script is already in the container

maingoh:
Close #186


dmake/dmake
```diff
@@ -57,6 +57,7 @@ try:
         app_args = 'app'
     elif command in ["shell"]:
         parser.add_argument("service", nargs='?', help="Run a shell session withing the docker base image for the given service. You may skip the app if there is no ambiguity, otherwise, you need to specify 'app/service'.")
+        parser.add_argument("-c", '--command', nargs='?', help="Run specified command in dmake shell instead of `docker.command` defined in `dmake.yml` (default: `bash`).")
```
thomas-riccardi: why simple quotes on `'--command'` ? for `"--dependencies"` it's double quotes

maingoh: Oups saw it too late

PR #193, opened 2018-05-02
Fixing installation issues

vdel:
- An error message was shown in case install.sh was run without argument
- DMAKE_NO_GPU was missing
- DMAKE_UID default has been changed to `""`
- Automatically leave DMAKE_SSH_KEY empty when no SSH key is found

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -121,17 +121,9 @@ if [ -f "${DMAKE_CONFIG_DIR}/config.sh" ]; then
 fi
 
 if [ -z "${DMAKE_SSH_KEY}" ]; then
-    declare -a KEYS=($(ls ${HOME}/.ssh/*.pub))
+    declare -a KEYS=($(ls ${HOME}/.ssh/*.pub 2> /dev/null))
     if [ -z "${KEYS}" ]; then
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -3,10 +3,14 @@
 set -e
 
 DMAKE_VERSION=0.1
+DMAKE_CONFIG_DIR=${DMAKE_CONFIG_DIR:-${HOME}/.dmake}
+DMAKE_UID=${DMAKE_UID:-}
```
thomas-riccardi: why change default to disabled?

install.sh
```diff
@@ -3,10 +3,14 @@
...
+DMAKE_PULL_CONFIG_DIR=${DMAKE_PULL_CONFIG_DIR:-1}
+DMAKE_NO_GPU=${DMAKE_NO_GPU:-1}
```
thomas-riccardi: why default to "allow run services that require gpu without gpu" ? I explicitly made this config a negative config: the default should be an error, and if the user is really OK to run services requiring a GPU without a GPU, then they can override by exporting `DMAKE_NO_GPU`

PR #209, opened 2018-05-25
Remove squashing of service image

### LINKED ISSUE #177: remove docker image squash: we have dockerfile support now if needed
remove docker image squash: we have dockerfile support now if needed

also, --squash was experimental and is deprecated (in favor of multi stage builds)

thomas-riccardi:
Closes #177 

It is a experimental feature, requiring manual docker configuration;
it has been deprecated in docker in favor of multi-stage builds.

Now that dmake supports multi-stage builds, we can safely remove this.


vdel:
Okay but we need to make sure the multi-stage build have two stages to force developers to extract artefacts

thomas-riccardi:
@vdel 
> Okay but we need to make sure the multi-stage build have two stages to force developers to extract artefacts

2 stages doesn't guarantee that developers don't leak anything secret. I don't thing it's worth trying to enforce 2 stages by code.

We need better documentation for on-premises images, that's all.

PR #213, opened 2018-05-25
Default to `--dependencies`, add `--no-dependencies`

### LINKED ISSUE #211: `-d` option should be the default behavior: run dependencies by default
*not* running dependencies by default mean that we often use external services instead of local ones: users could accidentally break things remotely without even noticing.

Running with dependencies by default would avoid that, and correspond to most use-cases.

thomas-riccardi:
Closes #211 

Also cleanup usage of common.options.dependencies.

### DIFF
--- a/dmake/dmake
+++ b/dmake/dmake
@@ -60,7 +60,7 @@ try:
         parser.add_argument("-c", '--command', nargs='?', help="Pass to `docker run` specified command instead of `docker.command` defined in `dmake.yml` (default: `bash`).")
         app_args = 'service'
     if command in ["shell", "run", "test"]:
-        parser.add_argument("-d", "--dependencies", required = False, default = False, action="store_true", help="By default, the service is launched as a standalone. If this flag is specified, the dependancies of the service and links specified in the test config as run as well.")
+        parser.add_argument("-d", "--dependencies", "--no-dependencies", required=False, default=True, dest='with_dependencies', action=common.BooleanAction, help="By default, the service is launched alongside its dependencies (service and link dependencies). These options control if dependencies are started.")
     if command in ["shell", "run"]:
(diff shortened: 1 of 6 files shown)


dmake/dmake
```diff
@@ -60,7 +60,7 @@ try:
         parser.add_argument("-c", '--command', nargs='?', help="Pass to `docker run` specified command instead of `docker.command` defined in `dmake.yml` (default: `bash`).")
         app_args = 'service'
     if command in ["shell", "run", "test"]:
-        parser.add_argument("-d", "--dependencies", required = False, default = False, action="store_true", help="By default, the service is launched as a standalone. If this flag is specified, the dependancies of the service and links specified in the test config as run as well.")
+        parser.add_argument("-d", "--dependencies", "--no-dependencies", required=False, default=True, dest='with_dependencies', action=common.BooleanAction, help="By default, the service is launched alongside its dependencies (service and link dependencies). These options control if dependencies are started.")
```
vdel: I would break retro-compatibility on-purpose here in order to no trick developers with a behavior they do not expect.
I would use the `--standalone` flag (and maybe `-a` ?)

thomas-riccardi:
The goal is to make non-standalone the default, so the expectation of the
developers is broken there: by default it will run with dependencies,
contrary to before.
Adding `--standalone` instead would not change that, would it?
And `--dependencies` & `--no-dependencies` is not tricky: it really does
what it says.

The idea was to later be able to start some of the dependencies, so
`--dependencies` would change from a boolean flag to a list of services.


On Tue, May 29, 2018 at 10:21 AM, Vincent Delaitre <notifications@github.com

thomas-riccardi:
Closing in favor of #334, now that we had other needs around dependencies.

PR #216, opened 2018-05-27
Add support for gdb in dmake shell

vdel:
Otherwise, GDB cannot access the trace.


dmake/deepobuild.py
```diff
@@ -1473,6 +1473,7 @@ def generate_shell(self, commands, service_name, docker_links, command=None):
         docker_opts += service.tests.get_mounts_opt(service_name, env)
 
         docker_base_image = self.docker.get_docker_base_image(service.get_base_image_variant())
+        docker_opts += """ --security-opt="apparmor=unconfined" --cap-add=SYS_PTRACE"""
```
thomas-riccardi: what about non-ubuntu-based hosts? (it seems COS also uses apparmor, so: rhel).

thomas-riccardi: Not our problem for now, opened #225

vdel: yep, I think most of dmake assumes ubuntu

PR #222, opened 2018-05-28
Improve pep8 formatting

vdel:
No changes appart from formatting


thomas-riccardi:
To avoid having to do that regularly we should enable pep8 checks during build.

dmake doesn't have the notion of source tests, currently the tests are only runtime tests executed on built docker image. This could work as we ship dmake sources in the image, but could be improved, cf #223

thomas-riccardi:
https://github.com/hhatto/autopep8 or more extreem https://github.com/google/yapf could do the job, with a manual exclution of re-indentation for some vertical alignment

vdel:
I see you point that we should enforce pep8 with a tool. Now this is more work that what I have done here and basically my editor is covered with annoying pep8 highlight so it would have been cool to merge.

I will just close this as it is now conflicting and I do not have the time to fix it. Feel free to re-open it and finish the work if you feel like it. ;-)

PR #237, opened 2018-06-07
Harmonize behaviour of repo digest computation between Linux and MacOS

### LINKED ISSUE #236: New digest breaks on MacOs
The digest function is still broken for two reasons:

1) #230 introduced a new hashing method that do not work on MacOS.

The reason is due to the output of the md5 function:

On Mac, `find . -type f -print0 | xargs -0 md5 -r` outputs this:
`MD5 (./MANIFEST.in) = 0f57c392c6c168939833a3e893068e37`

On Linux: `find . -type f -print0 | xargs -0 md5sum` outputs this:
`0f57c392c6c168939833a3e893068e37  ./MANIFEST.in`

The solution is to use `md5 -r` on MacOs but it only separates the hash and the path with one space (instead of two for Linux) so we must account for that too

2) For submodules, the `.git` file can point to the parent git repo with an absolute path that may differ from one computer to another. I propose to ignore the `.git` files for now.

vdel:
Closes #236

### DIFF
--- a/dmake/utils/dmake_md5
+++ b/dmake/utils/dmake_md5
@@ -25,9 +25,10 @@ set -e
 TARGET=$1
 VERSION=${2:-1}
 
-function md5sum_postproc {
+function md5sum_normalize_output {
     if [ $(uname) = "Darwin" ]; then
-        sed -e "s/\([a-f0-9]*\) /\1  /"
+        # this is necessary to return the same output as coreutils md5sum: 2 spaces between hash and path
+        sed -e 's/ /  /'
     else
         cat -
     fi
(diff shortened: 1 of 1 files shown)


dmake/utils/dmake_md5
```diff
@@ -25,10 +25,19 @@ set -e
 TARGET=$1
 VERSION=${2:-1}
 
+function md5sum_postproc {
+    if [ $(uname) = "Darwin" ]; then
+        sed -e "s/\([a-f0-9]*\) /\1  /"
```
thomas-riccardi: could document that we return same output as coreutils md5sum: 2 spaces.

Issue #243, opened 2018-06-28
`dmake test -d '*'`: unstable order between `test` and `run` for dependency services


thomas-riccardi:
# Scenario
Using `dmake-tutorial`: `dmake test -d '*'`:

## Run 1: OK
```
Here is the plan:
## Building Base ##
- base @ dmake-tutorial-worker-base::base
- base @ dmake-tutorial-web-base::base
## Building App ##
- build_docker @ dmake-tutorial/worker
- build_docker @ dmake-tutorial/web
## Testing App ##
- run_link @ links/dmake-tutorial/rabbitmq
- test @ dmake-tutorial/worker
- run @ dmake-tutorial/worker
- test @ dmake-tutorial/web
```

## Run 2: KO
```
Here is the plan:
## Building Base ##
- base @ dmake-tutorial-worker-base::base
- base @ dmake-tutorial-web-base::base
## Building App ##
- build_docker @ dmake-tutorial/worker
- build_docker @ dmake-tutorial/web
## Testing App ##
- run_link @ links/dmake-tutorial/rabbitmq
- run @ dmake-tutorial/worker
- test @ dmake-tutorial/worker
- test @ dmake-tutorial/web
```

# Expected result:
`run @ dmake-tutorial/worker` should always be executed *after* `test @ dmake-tutorial/worker`.

# Actual result:
The order is not stable, and regularly the `run` step is executed before `test`.

This is problematic for 2 reasons:
- we use services as dependencies for tests *before* testing them: this can lead to wrongly thinking a bug is in the parent service but is really in the dependency service
- `run` only starts a new container in the background, which may conflict with `test`, effectively run in parallel.

thomas-riccardi:
Real fix needs more control over the DAG, and choice from the dmake user: 

# auto detect what to test
- source for git diff?
- test service when it changes
- test parent services when a service changes
# test everything: `dmake test -d '*'`
# test one service only:
- runtime: optionally run its dependencies: `-d`
  - without its runtime dependencies (rely on remote, dev environment)
  - runtime: with its runtime dependencies
-  test: optionally also test its dependencies: `--test-dependencies` ?

Currently with the quickfix from #245 we test all dependencies.

PR #245, opened 2018-07-02
Quickfix: in test mode, always test dependencies before running them

thomas-riccardi:
Related to #243

Previously with `dmake test -d '*'` we did not guarantee steps order
between running a service as a dependency of another, and testing that
service.
This resulted in unwanted parallel execution of tests and run as
dependency.

Quickfix: always force testing a service before running it in test
mode (commands `test` and `deploy`).

Drawbacks:
- hardcoded `test` mode, we could use the dependency DAG to know if we
are in test mode instead
- when testing a service instead of `*`, we now also test its
dependencies services.


thomas-riccardi:
It's already successfully used on jenkins since last Thursday.

PR #252, opened 2018-07-09
Allow relative host paths in data_volumes

thomas-riccardi:
We currently assume paths are relative if they start with a dot: `.`
(we already used this convention to know the url uses the `file`
scheme)


thomas-riccardi:
In fact not needed, it is also a bad practice: if the file is needed just add it to the docker image.

PR #262, opened 2018-08-20
Add variable expansion for SharedVolumeMount `target`

### LINKED ISSUE #259: Allow variables in volume field
I want to use the named volume feature which seem to work only with `volumes` field (not `data_volumes`). I would like that the container path is an env variable, but dmake evaluate it after the check that it starts with a `/` to be an absolute path : 
```
# this doesn't work
volumes:
    - mynamedvolume: ${MY_DIR}
```

`data_volumes` allows to use env variables but doens't seem to handle named volumes ? I need both :)

Also maybe we should merge those two fields ?

thomas-riccardi:
Closes #259 

Variables are expanded as usual from the `env` related to the service
using the shared volume mount.

`target` validation (must be a full path) is now partially done later
as it must done after variable expansion, which cannot be done during
early yaml validation.


thomas-riccardi:
@maingoh does it fix your issue #259 ?

PR #273, opened 2018-09-11
Fix tests commands escaping: should be passed as-is to container

### LINKED ISSUE #272: commands in tests.commands is bash-evaluated too early
```
...
env:
  default:
    URL: http://foo

services:
  - service_name: e2e
...
    tests:
      commands:
        - curl -sSf "${URL}/api"
```
Results in this command being run in `docker run`: 
```
bash -c "curl -sSf \"/api\""
```
=> the `${URL}` part seems to be evaluated before reaching the container...

thomas-riccardi:
Closes #272.

Using simple quotes to wrap the commands we guarantee it will arrive
as-is in the `docker run` command.

### DIFF
--- a/dmake/common.py
+++ b/dmake/common.py
@@ -111,6 +111,10 @@ def escape_cmd(cmd):
 def wrap_cmd(cmd):
     return '"%s"' % cmd.replace('"', '\\"')
 
+def wrap_cmd_simple_quotes(cmd):
+    # Example: foo'bar -> 'foo'\''bar' (i.e. 3 concatenated literal strings: 'foo', \' and 'bar', which is interpreted by bash as one arg: foo'bar)
+    return "'%s'" % cmd.replace("'", "'\\''")
+
 def eval_str_in_env(value, env=None, strict=False, source=None):
     if env is None:
         env = {}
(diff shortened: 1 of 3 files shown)


dmake/common.py
```diff
@@ -111,6 +111,10 @@ def escape_cmd(cmd):
 def wrap_cmd(cmd):
     return '"%s"' % cmd.replace('"', '\\"')
 
+def wrap_cmd_simple_quotes(cmd):
+    # Example: foo'bar -> 'foo'\''bar' (i.e. 3 concatenated literal strings: 'foo', \' and 'bar', which is interpreted by bash as one arg: foo'bar)
```
maingoh: I am not sure I understand the comment with the code from the line just below. You are supposed to add two backslash not one ?

Also the escaping looks weird ? Replacing one `'` by a whole `'\\''` ?

thomas-riccardi: The comment explains what happens on the returned string; the code is more complicated because of python own escaping...

In bash double-quote strings is easy: just replace internal double quotes by backslash+double quote.
For simple quote strings it's not as simple: there is no backslash escaping for simple quote strings: `'foo\'bar'` doesn't work, it's `foo\` (single-quote delimited) then `bar` (undelimited) then opening of single-quote delimited..
The workaround is to use `\'` outside of quoted string, as I tried to explain in the comment...

maingoh: Alright, perfect !

PR #274, opened 2018-09-11
Add `env_exports` to `needed_services`

### LINKED ISSUE #271: Add env_exports for needed_services
For `needed_links` we can inject env vars to the services using them by adding `env_exports` on the `docker_links` definition (alongside `env` that overrides the runtime env for the docker link itself).

We would like to have that for `needed_services` too. Ideally we should have both `env_exports` and `env` at the service definition level, but we currently have neither (`env` at service level is a feature tracked here #159).
What we can easily do however is to add `env_exports` at the `needed_services` level: when defining how we want the dependency service.

Example:
```
  - service_name: e2e
    needed_services:
     - service_name: web
       link_name: web
       env_exports:
         WEB_URL: http://web:8000
```

This way we can still later add back `env` and `env_from` at the service definition level if needed.

thomas-riccardi:
Closes #271 

This is only used by the parent service when constructing its runtime
environment.
Thus, it does not change how the needed service should be started,
thus we don't need to distinguish it from other instances of
NeededService to build the global commands DAG: just changed `__str__`
and `__repr__`, not `__ne__` nor `__hash__`.

### DIFF
--- a/dmake/deepobuild.py
+++ b/dmake/deepobuild.py
@@ -100,7 +100,7 @@ class EnvBranchSerializer(YAML2PipelineSerializer):
     source    = FieldSerializer('string', optional=True, help_text='Source a bash file which defines the environment variables before evaluating the strings of environment variables passed in the *variables* field. It might contain environment variables itself.')
     variables = FieldSerializer('dict', child="string", default={}, help_text="Defines environment variables used for the services declared in this file. You might use pre-defined environment variables (or variables sourced from the file defined in the *source* field).", example={'ENV_TYPE': 'dev'})
 
-    def get_replaced_variables(self, additional_variables=None, docker_links=None, needed_links=None):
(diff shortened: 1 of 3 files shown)


dmake/deepobuild.py
```diff
@@ -1152,6 +1157,7 @@ class NeededServiceSerializer(YAML2PipelineSerializer):
     service_name    = FieldSerializer("string", help_text = "The name of the needed application part.", example = "worker-nn", no_slash_no_space = True)
     link_name       = FieldSerializer("string", optional = True, example = "worker-nn", help_text = "Link name.")
     env             = FieldSerializer("dict", child = "string", optional = True, default = {}, help_text = "List of environment variables that will be set when executing the needed service.", example = {'CNN_ID': '2'})
+    env_exports     = FieldSerializer("dict", child = "string", default = {}, help_text = "A set of environment variables that will be exported in services that use this service when testing.")
```
maingoh: no need for optional parameter just like env ?

Issue #278, opened 2018-09-12
DMAKE_USE_HOST_PORTS + host_port + services started multiple times by dependency


thomas-riccardi:
By dependency (notably for tests) we sometimes need to have multiple instances of a service executed at the same time. This breaks as soon as `DMAKE_USE_HOST_PORTS=1` (default) and a `host_port` is defined.

Maybe we could allow no `host_port` for such service (current `service` `port`s require both `container_port` and `host_port`).

Maybe we could ignore `host_port` for services started as dependencies of another service? Not really sure about that one.

PR #279, opened 2018-09-12
Allow no host_port on service ports: use random host port as if DMAKE_USE_HOST_PORTS=0

thomas-riccardi:
Related to #278

### DIFF
--- a/test/web/dmake.yml
+++ b/test/web/dmake.yml
@@ -45,7 +45,7 @@ services:
         start_script: deploy/start.sh
       ports:
         - container_port: 8000
-          host_port: 8000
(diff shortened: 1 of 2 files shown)


test/web/dmake.yml
```diff
@@ -77,3 +77,23 @@ services:
         directory: cover
         title: Web HTML coverage report
       cobertura_report: coverage.xml
+
+  - service_name: test-web2
+    # needed_links:
```
maingoh: remove ?

thomas-riccardi: good catch

PR #282, opened 2018-09-18
Add REPO_SANITIZED and BRANCH_SANITIZED host environment variables

thomas-riccardi:


dmake/test_common.py
```diff
@@ -0,0 +1,19 @@
...
+        ('test', 'test'),
+        ('Test', 'test'),
+        ('foo_bar', 'foo-bar'),
+        ('Foo_BAR', 'foo-bar'),
+        ('-foo', 'foo'),
+        ('foo__bar', 'foo-bar'),
+        ('dev-#123', 'dev--123'),
+        ('foo/bar', 'foo-bar'),
+    ]
+    for input, expected in tests:
```
maingoh: Nit: input is a global function in python

PR #310, opened 2018-11-26
Expose DMAKE_NO_DOCKER_CLEANUP to avoid dmake to clean docker resources

maingoh:


dmake/core.py
```diff
@@ -877,6 +877,8 @@ def make(options):
     # If on local, run the commands
     if common.is_local:
         result = os.system('bash %s' % file_to_generate)
+        if os.getenv('DMAKE_NO_DOCKER_CLEANUP') == '1':
+            return
```
thomas-riccardi: it is still useful to return the proper status code, and maybe also cleanup tmp files. The docker cleanup skipping should probably be done later, in `dmake/utils/dmake_remove_docker_containers_and_images`

dmake/utils/dmake_remove_docker_containers_and_images
```diff
@@ -18,6 +18,10 @@ set -e
 
 TMP_DIR=$1
 
+if [ "${DMAKE_NO_DOCKER_CLEANUP}" = "1" ]; then
+    exit 0;
```
thomas-riccardi: maybe print something to explain what happens ?
> Won't delete docker resources, as asked by `DMAKE_NO_DOCKER_CLEANUP=1`

PR #315, opened 2018-12-20
Allow `.env` variables in service build args and labels values substitution

thomas-riccardi:
It's useful to inject configs from env when the docker image is just
static assets (generated by angular for example).


thomas-riccardi:
TODO: dmake tests

thomas-riccardi:
We should in fact not encourage dmake users to build differentiated images per environment: we tend to have a unique image that is tested on several environments before being deployed to production.

PR #316, opened 2018-12-22
Fix release process

vdel:
I realized I had also fixed the release process (@maingoh)
Feel free to close it if you do not want/need to merge.

### DIFF
--- a/dmake/commands/release.py
+++ b/dmake/commands/release.py
@@ -51,7 +51,7 @@ def entry_point(options):
         questions = [
             inquirer.List(
                 'release_tag',
-                message="Here are only the latest tags per major-minor version. Which tag do you want to release?",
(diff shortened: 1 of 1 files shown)


dmake/commands/release.py
```diff
@@ -51,7 +51,7 @@ def entry_point(options):
         questions = [
             inquirer.List(
                 'release_tag',
-                message="Here are only the latest tags per major-minor version. Which tag do you want to release?",
+                message="Here are only the latest tags per major-minor version. Which is the target tag to release on?",
```
maingoh: I am not sure this sentence is better ^^. It sounds less english, I think you need a name after `which` in this context.

vdel:
Yes I have seen you had done some fixes. I just had those few changes in addition. Btw, I fixed the sentences. Feel free to merge

PR #327, opened 2019-01-14
Use official registry URL

thomas-riccardi:
`index.docker.io` was the original URL (registry v1). Since v2 the
official registry URL is `registry-1.docker.io`:
https://github.com/moby/moby/blob/a3f626d1019bd049dcc46ac8f64fe08b057e89f9/registry/config.go#L49-L52

And for retro-compatibility the client config.json file auth section
still uses `https://index.docker.io/v1/` as key.

Also raise a more complete exception when the registry returns 404.

### DIFF
--- a/dmake/docker_config.py
+++ b/dmake/docker_config.py
@@ -56,6 +56,9 @@ def get_docker_config_auth(registry_url, dockercfg = '~/.docker/config.json'):
         if registry.startswith(registry_url):
             return credentials_store, registry, registry_data
     else:
+        if registry_url == 'https://registry-1.docker.io':
+            # fallback to default v1 creds
+            return get_docker_config_auth('https://index.docker.io', dockercfg)
         raise DMakeException('Auth not found for registry %s in docker config file %s' % (registry_url, dockercfg))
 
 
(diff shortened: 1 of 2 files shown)


dmake/docker_config.py
```diff
@@ -56,6 +56,9 @@ def get_docker_config_auth(registry_url, dockercfg = '~/.docker/config.json'):
         if registry.startswith(registry_url):
             return credentials_store, registry, registry_data
     else:
+        if registry_url == 'https://registry-1.docker.io':
+            # fallback to default v1 creds
+            return get_docker_config_auth('https://index.docker.io', dockercfg)
```
maingoh: not `https://index.docker.io/v1/` ?

thomas-riccardi: no, the `/v1` suffix is a hack on docker `config.json` side, and we already supported it by searching for a prefix string instead of string equality (see https://github.com/Deepomatic/dmake/blob/02c517ee165b04aeadaaf83ffd87ba8d8aab53bc/dmake/docker_config.py#L48)

This fallback effectively get auth exactly like it was before this PR (we always get into this fallback mode, but I though it would be cleaner to do it as a fallback than as a special case directly at the beginning of the function. In fact it doesn't really matter).

Issue #235, opened 2018-06-06
Deployment steps order is random; it should respect services dependencies instead


thomas-riccardi:
Some deployments require ordered deployment between the services: Service B needs Service A already deployed so its own deployment works.

(For example nginx resolves its upstreams hostnames at startup and returns an error if hostname doesn't exists. For a first deployment in kubernetes the service domain name doesn't exists yet.)

dmake steps orders inside of stages seems to be random (and even changes on each run with python3; with python2 the order is stable across runs).

# TODO
- [ ] remove unstable order in python3
- [ ] sort deployments steps according to service dependencies

thomas-riccardi:
Related to #243:
To fix the deployment order issue the easy way is to force deployment dependency like we did in #243 quickfix.
It will have the same limitations as for #243: too many deployments (always deploy all dependencies).

The real fix is the same as for #243

thomas-riccardi:
In fact the services deployments order is correct when executing `dmake deploy '*'` with tests, because the tests force the dependency at the test/run level/stages, and thus the steps depths on top of `test` (i.e. `deploy`) are also correctly ordered.
This is a side effect of #245, and doesn't work if we disable the tests (`DMAKE_SKIP_TESTS=1 dmake deploy '*'`).

thomas-riccardi:
Here are all the possibilities for a service `A`, with some of them already implemented:
- `test` `A`:
  - just test `A`: DONE: `dmake test A`
  - also run dependencies
    - recursively?
  - also test dependencies
    - recursively? : DONE: `dmake test -d A`
- `deploy` `A`:
  - just deploy `A`: DONE: `DMAKE_SKIP_TESTS=1 dmake deploy A`
  - deploy `A` + test `A` (with all the combinations of `test` `A`, see below): 
    - deploy `A` + test `A` recursively: DONE `dmake deploy A`
  - also deploy dependencies:
    - just also deploy dependencies
      - recursively?
    - deploy + test dependencies
      - recursively?

thomas-riccardi:
After analyzing the dmake usage: for now let's:
- implement the quickfix for deploy as done in #245 
- support `--no-dependencies` to be able to only deploy one service as before (`DMAKE_SKIP_TESTS=1 dmake deploy A --no-dependencies`)
- it's probably the occasion to switch `--dependencies` by default in all commands, cf #213

thomas-riccardi:
Closing via #338 and #334.

Followup in #342 .

[...]

PR #334, opened 2019-01-23
Default to --dependencies & add cli options --no-dependencies, --standalone, -s

### LINKED ISSUE #211: `-d` option should be the default behavior: run dependencies by default

thomas-riccardi:
Closes #211, prerequisite for #235.

* Support cli options --no-dependencies, --standalone, -s
* Now default to `--dependencies` for dmake shell, run, test, deploy
It will avoid accidentally using remote resources (such as remote
database): use `--standalone` or `-s` to force disabling local
dependencies execution.
Also, now, `dmake deploy` can be executed without dependencie
* Update the README with the breaking change around `--dependencies`
* Update the Jenkinsfiles to use new --dependencies & --standalone

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -40,10 +40,7 @@ node {
 
   def self_test = (params.REPO_TO_TEST == 'deepomatic/dmake')
 
-  def dmake_with_dependencies = ''
-  if (params.DMAKE_WITH_DEPENDENCIES) {
-    dmake_with_dependencies = '--dependencies'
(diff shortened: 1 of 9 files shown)


thomas-riccardi:
Still missing Jenkinsfile changes.

PR #365, opened 2019-04-03
Add kubernetes multi `manifests`

thomas-riccardi:
Old mono `manifest` and multi `manifests` are exclusive.

Also add some tests for kubernetes deployments.

For now it just tests dmake.yml parsing, as dmake tests don't test
`deploy`.


dmake/deepobuild.py
```diff
@@ -755,9 +755,16 @@ class KubernetesDeploySerializer(YAML2PipelineSerializer):
     context     = FieldSerializer("string", help_text="kubectl context to use.")
     namespace   = FieldSerializer("string", optional=True, help_text="Kubernetes namespace to target (overrides kubectl context default namespace")
     manifest    = FieldSerializer(KubernetesManifestSerializer(), optional=True, help_text="Kubernetes manifest defining all the resources needed to deploy the service")
```
maingoh: We can also break the compatibility and force the old field to be updated (I guess it only affects one repository ?)

dmake/deepobuild.py
```diff
@@ -804,8 +811,12 @@ def generate_and_write_additional_resources(sources, filename):
         # copy/render template manifest file
         context = common.eval_str_in_env(self.context, env)
         namespace = common.eval_str_in_env(self.namespace, env) if self.namespace else ""
-        if self.manifest is not None:
-            user_manifest_filename = 'kubernetes-user-manifest.yaml'
+        manifests = []
+        if self.manifest:
+            manifests.append(self.manifest)
+        manifests.extend(self.manifests)
+        for n, manifest in enumerate(manifests, start=1):
```
maingoh: why ignoring the first one ? maybe add a comment to explain why

thomas-riccardi: I don't ignore the first item, I instead start the `n` enumeration to `1`; to have fancy filenames.
Maybe it's not a good idea though..

dmake/deepobuild.py
thomas-riccardi: yes we could; but it's too much work for now. I still forbid using both `manifest` and `manifests` at the same time, so we can still do it later relatively easily.

PR #366, opened 2019-04-10
Deprecate docker links: services can fully replace them

vdel:
Docker links duplicate many notions and are not necessary. In order to unify how we handle the various concepts in a deployment pipeline, let's get rid of them.

### Migration indications:

For each link, create a service. Use the following mapping to update the service definition:

| Before       |     After    |
| :------------ | :------------- |
| `docker_links`::`image_name`      | `services`::`config`::`docker_image` |
| `docker_links`::`link_name`       | `services`::`service_name` |
| `docker_links`::`probe_ports`     | `services`::`config`::`probe_ports` |
| `docker_links`::`env`             | `services`::`config`::`env_override` |
| `docker_links`::`volumes`         | `services`::`config`::`volumes` |
| `docker_links`::`need_gpu`        | `services`::`config`::`need_gpu` |
| `docker_links`::`testing_options` | `services`::`config`::`docker_opts` |
| `docker_links`::`env_exports`     | `services`::`needed_services`::`env_exports` |

I'm not sure the last one will work, why is it injected when calling needed_services ? Anyway, I haven't changed this behavior here.

### DIFF
--- a/dmake/deepobuild.py
+++ b/dmake/deepobuild.py
@@ -6,7 +6,7 @@ import importlib
 import re
 import requests.exceptions
 from string import Template
-from dmake.serializer import ValidationError, FieldSerializer, YAML2PipelineSerializer
+from dmake.serializer import ValidationError, FieldSerializer, YAML2PipelineSerializer, SerializerMixin
 import dmake.common as common
 from dmake.common import DMakeException, SharedVolumeNotFoundException, append_command
 import dmake.kubernetes as k8s_utils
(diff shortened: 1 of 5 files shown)


dmake/deepobuild.py
```diff
@@ -1054,6 +1056,14 @@ def _validate_(self, file, needed_migrations, data, field_name):
         if common.is_string(data):
             data = {'service_name': data}
         result = super(NeededServiceSerializer, self)._validate_(file, needed_migrations=needed_migrations, data=data, field_name=field_name)
+
+        # We link the service by default
+        if self.link_name is None:
+            # We link the service using its name, unless the service name has ':' in which case it is a variant
+            # and it is better to let the user chose the link name if needed
+            if self.service_name.find(':') < 0:
+                self.__fields__['link_name'] = self.service_name
```
thomas-riccardi: if we keep this, we should document it in the field description

thomas-riccardi: it will possibly lead to silent link name conflict in case of multiple needed services on same service but with different env customization.
=> I suggest to *not* do it then.
=> maybe also add an explicit check on link name duplication somewhere?

dmake/utils/dmake_wait_for_probe_ports
```diff
@@ -0,0 +1,32 @@
...
+
+if [ "${PROBE_PORTS}" = "none" ]; then
+    :
+else
+    if [ "${PROBE_PORTS}" = "auto" ]; then
+        PROBE_PORTS=$(docker ps -f id=${CONTAINER_ID} --format "{{.Ports}}" | sed "s/ *//g" | sed "s/\([0-9]*\.\)\{3\}[0-9]*:[0-9]*->//g")
+    fi
+    if [ ! -z "${PROBE_PORTS}" ]; then
+        ROOT=$(dirname $0)
+        >&2 docker run --rm --link ${CONTAINER_ID}:link_to_wait -v ${ROOT}/dmake_wait_for_it:/usr/bin/dmake_wait_for_it -v ${ROOT}/dmake_wait_for_it_wrapper:/usr/bin/dmake_wait_for_it_wrapper -i ubuntu dmake_wait_for_it_wrapper "link_to_wait" "${PROBE_PORTS}"
```
thomas-riccardi: nit: logs could be more informative with the real link name instead of the hardcoded `link_to_wait`, even if it requires passing it as an extra argument

dmake/utils/dmake_wait_for_probe_ports
```diff
@@ -0,0 +1,32 @@
...
+
+if [ $# -ne 2 ]; then
+    >&2 echo -e "$0: Missing arguments"
+    exit 1
+fi
+
+set -e
+
+CONTAINER_ID=${1}
+PROBE_PORTS=${2}
```
thomas-riccardi: nit: the `{}` are not necessary, and are unusual for positional arguments

dmake/deepobuild.py
```diff
@@ -834,7 +836,7 @@ def get_cmd(self):
         cmd = common.escape_cmd(cmd)
         return 'bash -c "%s"' % cmd
 
-class DeployConfigSerializer(YAML2PipelineSerializer):
+class DeployConfigSerializer(YAML2PipelineSerializer, ProbePortMixin):
```
thomas-riccardi: This enables the `auto` port probing for all Services: this could break things.

For example `test-gpu` root image declare some ports, which are now auto waited-for when `test-gpu` is run ('dmake run test-gpu`, or if `test-gpu` is ever added as a `needed_services`).

thomas-riccardi:
One more issue: `docker_links` host volumes are sometimes ignored (host persistence for developers, no persistence for tests/Jenkins): https://github.com/Deepomatic/dmake/blob/8d3134bc12ea38a31e6aef50e44d7e2e4b6abcc8/dmake/deepobuild.py#L413-L417
This feature is missing from Services.

PR #367, opened 2019-04-11
Shows the real battle plan

vdel:
When starting dmake, it's show the list of action it will follow.
Some actions turn to however be empty, especially when no tests are set for a service, or when you use an external image for a service (see #368, you do not expect a build step) which is a bit confusing for the user. This PR allow to show the "real" battle plan, once empty steps have been filtered out.

### DIFF
--- a/dmake/core.py
+++ b/dmake/core.py
@@ -779,19 +779,6 @@ def make(options, parse_files_only=False):
         if not common.is_pr:
             ordered_build_files.append(('Deploying', list(deploy)))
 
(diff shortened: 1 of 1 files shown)


thomas-riccardi:
One extra case is the deployment steps: when calling `dmake deploy` on a branch with no deployment step configured, it still prints the `deploy` step, but it's normal: it still pushes the docker images.
Ideally the docker push should be more explicit and not hidden behind `deploy`.

PR #368, opened 2019-04-11
Allow to define services that use external images

vdel:
One may want to use external images in a project, e.g. the pre-built Deepomatic neural worker version.
You may be tempted to use dmake links to define this service, the problem is that you need to link it to RabbitMQ to communicate with it, and there is no way to do that with dmake links.

So you end up using a service (that you can link to a link), but there is no way to use an external image for a service, you have to build a Docker image. So this PR allows to do it.

### DIFF
--- a/dmake/core.py
+++ b/dmake/core.py
@@ -5,7 +5,7 @@ import uuid
 
 import dmake.common as common
 from dmake.common import DMakeException, SharedVolumeNotFoundException
-from dmake.deepobuild import DMakeFile, append_command
+from dmake.deepobuild import DMakeFile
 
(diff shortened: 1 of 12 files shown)


dmake/core.py
```diff
@@ -5,7 +5,7 @@
 
 import dmake.common as common
 from dmake.common import DMakeException, SharedVolumeNotFoundException
-from dmake.deepobuild import DMakeFile, append_command
```
thomas-riccardi: I believe both PR and code would be more readable with from `dmake.common import append_command`, as you did in `dmake/docker_image.py`

thomas-riccardi: I pushed a commit doing that systematically, PR is much more readable

PR #382, opened 2019-05-20
Using withEnv instead of env in generated DMakeFile (issue #381)

### LINKED ISSUE #381: Replace Using env variables has side effects when using multiple nodes
To avoid problems of side effects after changing env, especially using multiple nodes, it is better to set a temporary context.

One safe way to alter the environment is:
```
 withEnv(['REPO=vulcan',
          'COMMIT=32b6a9b9f19d0a5ca0adcc1859f93c508ea3040e',
          'BUILD=65',
          'BRANCH=dev-jenkins-up2-slave',
          'DMAKE_TMP_DIR=/tmp/dmake_tmp_1558351859164023995_32383/'
    ]) {
         // Something using the env here

 }
```
This approach does not poison the env after the command execution.

ibreschi:
Closes #381 

**Before**
```
  env.REPO = "vulcan"
  env.COMMIT = "0c226d25bcf84dc75d2019c3d5348e6f66c274e3"
  env.BUILD = "83"
  env.BRANCH = "dev-jenkins-up2-slave"
  env.DMAKE_TMP_DIR = "/tmp/dmake_tmp_1558452978304966359_14391/"
...
...
```
**After**
```
withEnv([
  "REPO=vulcan",
  "COMMIT=0c226d25bcf84dc75d2019c3d5348e6f66c274e3",
  "BUILD=83",
  "BRANCH=dev-jenkins-up2-slave",
  "DMAKE_TMP_DIR=/tmp/dmake_tmp_1558452984505488394_6894/"]) {
....
....
}
```

### DIFF
--- a/dmake/core.py
+++ b/dmake/core.py
@@ -425,14 +425,18 @@ def generate_command_pipeline(file, cmds):
 
     if common.build_description is not None:
         write_line("currentBuild.description = '%s'" % common.build_description.replace("'", "\\'"))
-    write_line('try {')
-    indent_level += 1
 
     cobertura_tests_results_dir = os.path.join(common.relative_cache_dir, 'cobertura_tests_results')
     emit_cobertura = False
 
     for cmd, kwargs in cmds:
-        if cmd == "stage":
+        if cmd == "try":
+            write_line('try {')
+            indent_level += 1
+        elif cmd == "try_end":
+            indent_level -= 1
+            write_line('}')
+        elif cmd == "stage":
             name = kwargs['name'].replace("'", "\\'")
             write_line('')
             if kwargs['concurrency'] is not None and kwargs['concurrency'] > 1:
@@ -481,8 +485,17 @@ def generate_command_pipeline(file, cmds):
             write_line("env.%s = readFile '%s'" % (kwargs['var'], file_output))
             if kwargs['fail_if_empty']:
                 write_line("sh('if [ -z \"${%s}\" ]; then exit 1; fi')" % kwargs['var'])
-        elif cmd == "env":
-            write_line('env.%s = "%s"' % (kwargs['var'], kwargs['value']))
+        elif cmd == "with_env":
+            write_line('withEnv([')
+            indent_level += 1
+            environment = kwargs['value']
+            for element in environment[:-1]:
+                write_line('"%s=%s",' % (element[0], element[1]))
+            last_element = environment[-1]
+            write_line('"%s=%s"]) {' % (last_element[0], last_element[1]))
+        elif cmd == "with_env_end":
+            indent_level -= 1
+            write_line('} // WithEnv end')
         elif cmd == "git_tag":
             if common.repo_url is not None:
                 write_line("sh('git tag --force %s')" % kwargs['tag'])
@@ -548,14 +561,19 @@ def generate_command_pipeline(file, cmds):
     write_line('finally {')
     write_line('  sh("dmake_clean")')
     write_line('}')
-
+    indent_level -= 1
+    write_line('} // WithEnv end')
 ###############################################################################
 
 def generate_command_bash(file, cmds):
     file.write('test "${DMAKE_DEBUG}" = "1" && set -x\n')
     file.write('set -e\n')
     for cmd, kwargs in cmds:
-        if cmd == "stage":
+        if cmd == "try":
+            pass
+        elif cmd == "try_end":
+            pass
+        elif cmd == "stage":
             file.write("\n")
             file.write("echo -e '\n## %s ##'\n" % kwargs['name'])
         elif cmd == "stage_end":
@@ -583,9 +601,13 @@ def generate_command_bash(file, cmds):
             file.write("%s=`%s`\n" % (kwargs['var'], kwargs['shell']))
             if kwargs['fail_if_empty']:
                 file.write("if [ -z \"${%s}\" ]; then exit 1; fi\n" % kwargs['var'])
-        elif cmd == "env":
-            file.write('%s="%s"\n' % (kwargs['var'], kwargs['value'].replace('"', '\\"')))
-            file.write('export %s\n' % kwargs['var'])
+        elif cmd == "with_env":
+            environment = kwargs['value']
+            for element in environment:
+                file.write('%s="%s"\n' % (element[0], element[1].replace('"', '\\"')))
+                file.write('export %s\n' % element[0])
+        elif cmd == "with_env_end":
+            pass
         elif cmd == "git_tag":
             file.write('git tag --force %s\n' % kwargs['tag'])
             file.write('git push --force %s refs/tags/%s || echo %s\n' % (common.remote, kwargs['tag'], tag_push_error_msg))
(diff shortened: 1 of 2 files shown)


dmake/core.py
```diff
@@ -788,11 +807,13 @@ def make(options, parse_files_only=False):
     # Generate the list of command to run
     common.logger.info("Generating commands...")
     all_commands = []
+    append_command(all_commands, 'begin_env')
     append_command(all_commands, 'env', var = "REPO", value = common.repo)
     append_command(all_commands, 'env', var = "COMMIT", value = common.commit_id)
     append_command(all_commands, 'env', var = "BUILD", value = common.build_id)
     append_command(all_commands, 'env', var = "BRANCH", value = common.branch)
-    append_command(all_commands, 'env', var = "DMAKE_TMP_DIR", value = common.tmp_dir)
+    append_command(all_commands, 'last_env', var = "DMAKE_TMP_DIR", value = common.tmp_dir)
```
thomas-riccardi: the api would be cleaner with 
`begin_env` command taking a list of key/value as argument (or kwargs directly maybe?)
it would avoid the hack `last_env`

dmake/core.py
```diff
@@ -788,11 +807,13 @@ def make(options, parse_files_only=False):
     # Generate the list of command to run
     common.logger.info("Generating commands...")
     all_commands = []
+    append_command(all_commands, 'begin_env')
```
thomas-riccardi: each command that creates a new nesting level/scope should have it's corresponding `end` command to close the scope.
Prior examples:
- `stage`: `stage_end`
- `lock`: `lock_end`
- `timeout`: `timeout_end`
=> I also suggest to name it `envs` and `envs_end` (or without the plural `s`, as you wish; or `with_env`? the advantage is that it's a new name: easier to find coding errors if we forgont one old `env`)

dmake/core.py
```diff
@@ -425,14 +425,15 @@ def write_line(data):
 
     if common.build_description is not None:
         write_line("currentBuild.description = '%s'" % common.build_description.replace("'", "\\'"))
-    write_line('try {')
```
thomas-riccardi: why make this explicit with a new command? it has no value AFAIK: dmake internally handles cleaning everything with this global try/catch, independently of the generated commands.

Also, if we would have kept the new command `try`, then we should add and emit the corresponding `try_end`.

dmake/core.py
```diff
@@ -481,8 +482,16 @@ def write_line(data):
             write_line("env.%s = readFile '%s'" % (kwargs['var'], file_output))
             if kwargs['fail_if_empty']:
                 write_line("sh('if [ -z \"${%s}\" ]; then exit 1; fi')" % kwargs['var'])
+        elif cmd == "begin_env":
+            write_line('withEnv([')
+            indent_level += 1
         elif cmd == "env":
-            write_line('env.%s = "%s"' % (kwargs['var'], kwargs['value']))
+            write_line('"%s=%s",' % (kwargs['var'], kwargs['value']))
```
thomas-riccardi: what about values escaping ? that being said, we probably had exactly the same issues with the previous implementation, so that should be OK for now.

dmake/core.py
```diff
@@ -425,14 +425,18 @@ def write_line(data):
...
         write_line("currentBuild.description = '%s'" % common.build_description.replace("'", "\\'"))
-    write_line('try {')
-    indent_level += 1
 
     cobertura_tests_results_dir = os.path.join(common.relative_cache_dir, 'cobertura_tests_results')
     emit_cobertura = False
 
     for cmd, kwargs in cmds:
-        if cmd == "stage":
+        if cmd == "try":
```
vdel: Why adding `try` support if not supported in bash ? Do we need it ?
This breaks the idea that dmake should behave the same on Jenkins and a developer machine !

dmake/core.py
```diff
@@ -481,8 +485,17 @@ def write_line(data):
             write_line("env.%s = readFile '%s'" % (kwargs['var'], file_output))
             if kwargs['fail_if_empty']:
                 write_line("sh('if [ -z \"${%s}\" ]; then exit 1; fi')" % kwargs['var'])
-        elif cmd == "env":
```
vdel: Why removing env support ?
When need some variables to be global!

dmake/core.py
```diff
@@ -788,11 +810,16 @@ def make(options, parse_files_only=False):
     # Generate the list of command to run
     common.logger.info("Generating commands...")
     all_commands = []
-    append_command(all_commands, 'env', var = "REPO", value = common.repo)
```
vdel: The 4 first variables need to be global.

dmake/core.py
```diff
@@ -788,11 +810,16 @@ def make(options, parse_files_only=False):
...
-    append_command(all_commands, 'env', var = "DMAKE_TMP_DIR", value = common.tmp_dir)
+
+    environment = [
+        ("REPO", common.repo),
+        ("COMMIT", common.commit_id),
+        ("BUILD", common.build_id),
+        ("BRANCH", common.branch),
+        ("DMAKE_TMP_DIR", common.tmp_dir)
+    ]
+    append_command(all_commands, 'with_env', value = environment)
```
vdel: I do not see the closing `with_env_end` so I guess there is a big design issue here.

dmake/core.py
```diff
@@ -788,11 +810,16 @@ def make(options, parse_files_only=False):
...
+    append_command(all_commands, 'try')
```
vdel: Why adding a try without control on the catch block ? It seems conceptually dangerous to wrap everything in a try: we will miss every error.
Again, there seems to be a big design issue here.

dmake/core.py
ibreschi: Given that envs are global, it's a little bit dangerous to keep the env command. People will think that it can be used as a normal scoped variable and we will have again multi arch problems. A solution in between would be to rename the command to global_env. What do you think?

dmake/core.py
ibreschi: I would not say they need to be, but yes they can be global.

dmake/core.py
ibreschi: The multi build obliged me to move the try catch inside a withEnv to set DMAKE_TMP_DIR. Without it the nodes will sync the variable and try to access a folder that does not exists. 
So I had to find a different way to write it. If you have suggestions I'm open to discuss.

dmake/core.py
ibreschi: Yes you are right. Adding withEnv properly (a scoped space  ) would need more work. I just added the two commands with_env and with_env_end. The close is done manually, but maybe I can find a place in the make method to put it. I'll do that

dmake/core.py
ibreschi: I did not want to dig to much in code generation. the try command is actually just an hack that I was forced to put as explained above.

dmake/core.py
```diff
@@ -425,14 +425,18 @@ def write_line(data):
 
     if common.build_description is not None:
         write_line("currentBuild.description = '%s'" % common.build_description.replace("'", "\\'"))
-    write_line('try {')
```
vdel: Just above this, you could do something like this:

```
    append_command(all_commands, 'global_env', prepend=True, var = "REPO", value = common.repo)
    append_command(all_commands, 'global_env', prepend=True, var = "COMMIT", value = common.commit_id)
    append_command(all_commands, 'global_env', prepend=True, var = "BUILD", value = common.build_id)
    append_command(all_commands, 'global_env', prepend=True, var = "BRANCH", value = common.branch)
    append_command(all_commands, 'with_env', prepend=True, value = [("DMAKE_TMP_DIR", common.tmp_dir)])
    append_command(all_commands, 'try', prepend=True)
```

ibreschi: Done

ibreschi:
Closed for inactivity

PR #384, opened 2019-05-21
Replacing temp file used to return variables (issue #383)

### LINKED ISSUE #383: Replacing temp file used to return variables
To return the DAEMON_ID,  we use to write to a file the output of dmake_run_docker_daemon  than read it back into env and than use it. 
We should replace this mechanism with a withEnv that is simpler and safer at the same time

ibreschi:
Closes #383 

**Before**
```
    sh('dmake_run_docker_daemon "vulcan" "vulcan/init-db---5691079760623584708" "" "" --env-file /tmp/dmake_tmp_1558450334962678058_23997/env.txt.d1d3df94-0780-49f9-aee7-9297d7c71556   $(dmake_return_docker_links vulcan postgres) -i init-db:dev-jenkins-up2-slave-82 > /var/jenkins_home/jobs/vulcan-test-multi-node/branches/dev-jenkins-up2-slave/workspace@2/.dmake/output_0dd6e73b-a517-4b02-a939-c926d94b2cc1')
    env.DAEMON_ID = readFile '/var/jenkins_home/jobs/vulcan-test-multi-node/branches/dev-jenkins-up2-slave/workspace@2/.dmake/output_0dd6e73b-a517-4b02-a939-c926d94b2cc1'
    sh("dmake_exec_docker \${DAEMON_ID} bash -c \"T=0; sleep 0; while [ 1 ]; do echo \\\"Running readiness probe\\\"; ls \\\"/tmp/init-db-ready\\\"; if [ \\\"\\\$?\\\" = \\\"0\\\" ]; then echo \\\"... ready\\\"; exit 0; fi; T=\\\$((T+5)); sleep 5; done; exit 1;\"")
```

**After**
```
    def DAEMON_ID = sh(script:'dmake_run_docker_daemon "vulcan" "vulcan/init-db---2965137994837651050" "" "" --env-file /tmp/dmake_tmp_1558450224598011660_11366/env.txt.99f44ac6-dd52-4269-9c91-5b4eafb3f729   $(dmake_return_docker_links vulcan postgres) -i init-db:dev-jenkins-up2-slave-82', returnStdout: true).trim()
    withEnv([
      "DAEMON_ID=${DAEMON_ID}"]) {
      sh("dmake_exec_docker \${DAEMON_ID} bash -c \"T=0; sleep 0; while [ 1 ]; do echo \\\"Running readiness probe\\\"; ls \\\"/tmp/init-db-ready\\\"; if [ \\\"\\\$?\\\" = \\\"0\\\" ]; then echo \\\"... ready\\\"; exit 0; fi; T=\\\$((T+5)); sleep 5; done; exit 1;\"")
    } // end env
```

### DIFF
--- a/dmake/core.py
+++ b/dmake/core.py
@@ -9,6 +9,8 @@ from dmake.deepobuild import DMakeFile
 
 tag_push_error_msg = "Unauthorized to push the current state of deployment to git server. If the repository belongs to you, please check that the credentials declared in the DMAKE_JENKINS_SSH_AGENT_CREDENTIALS and DMAKE_JENKINS_HTTP_CREDENTIALS allow you to write to the repository."
 
+# List that stores all the generated variables
+VARIABLES = []
 ###############################################################################
 
 def find_symlinked_directories():
@@ -479,6 +481,13 @@ def generate_command_pipeline(file, cmds):
                     commands_list.append("cmd%d: { sh('%s') }" % c)
                 write_line(','.join(commands_list))
                 write_line(')')
+        elif cmd == "assign_var":
+            name = kwargs['var']
+            if name not in VARIABLES:
+               write_line("def %s = sh(script:'%s', returnStdout: true).trim()" % (name, kwargs['shell']))
+               VARIABLES.append(name)
+            else:
+               write_line("%s = sh(script:'%s', returnStdout: true).trim()" % (name, kwargs['shell']))
         elif cmd == "read_sh":
             file_output = os.path.join(common.cache_dir, "output_%s" % uuid.uuid4())
             write_line("sh('%s > %s')" % (kwargs['shell'], file_output))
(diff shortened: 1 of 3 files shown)


dmake/core.py
```diff
@@ -479,6 +481,13 @@ def write_line(data):
                     commands_list.append("cmd%d: { sh('%s') }" % c)
                 write_line(','.join(commands_list))
                 write_line(')')
+        elif cmd == "assign_var":
+            name = kwargs['var']
+            if name not in VARIABLES:
```
vdel: I do not know the `def` syntax but I guess it has to be used exactly once per variable in every stage.

So what is done here will likely not work. Imagine you have two stages declaring the same variable.
When generating the second variable you will believe it has been already declared, which it won't because it was done in another stage.

To solve this I propose to suffix the variable name with a number to make it unique and to always use `def`. (We need to check it is unique).
Then, you can use the returned variable name in subsequent commands.

ibreschi: Yes you are right. If we accept that variable are global I can use avoid using def and this will solve all the problems.

vdel: looks good ! :)

ibreschi:
Closed for inactivity

PR #389, opened 2019-06-07
Optionally use telepresence for dmake shell

thomas-riccardi:
This is a test to use [telepresence](https://www.telepresence.io) for a more integrated `dmake shell` experience: local container execution in the full network context of the remote kubernetes deployment.

## Example
```
DMAKE_TELEPRESENCE=1 dmake shell -s foo
DMAKE_TELEPRESENCE=1 DMAKE_TELEPRESENCE_ARGS=--mount=false dmake shell -s foo
```

## TODO later
- only for dmake shell
- select correct kubernetes context according to deployment declared in dmake
- support `--swap-deployment` according to some config in dmake.yml

### DIFF
--- a/dmake/utils/dmake_run_docker
+++ b/dmake/utils/dmake_run_docker
@@ -48,8 +48,9 @@ if [ ! -z "${TMP_DIR}" ]; then
     echo ${NAME} >> ${TMP_DIR}/containers_to_remove.txt
 fi
 
+DOCKER_RUN_ARGS=( )
+
 # support some GPU modes using nvidia docker v2
-DOCKER_RUN_ARGS=( run )
 if [[ "${DMAKE_DOCKER_RUN_WITH_GPU}" == 'yes' ]]; then
   DEVICES=${DMAKE_GPU#GPU_}
   DOCKER_RUN_ARGS+=( --runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=${DEVICES:-all} )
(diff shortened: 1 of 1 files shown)


dmake/utils/dmake_run_docker
```diff
@@ -48,13 +48,21 @@ if [ ! -z "${TMP_DIR}" ]; then
...
 elif [[ "${DMAKE_DOCKER_RUN_WITH_GPU}" == 'none' ]]; then
   DOCKER_RUN_ARGS+=( -e NVIDIA_VISIBLE_DEVICES=none )
 fi
 
-docker "${DOCKER_RUN_ARGS[@]}" --name ${NAME} "$@"
+DOCKER_RUN_ARGS+=( --name ${NAME} "$@" )
+
+# support telepresence
+if [[ "${DMAKE_TELEPRESENCE}" == '1' ]]; then
+  telepresence "${DMAKE_TELEPRESENCE_ARGS[@]}" --docker-run "${DOCKER_RUN_ARGS[@]}"
```
maingoh: `DMAKE_TELEPRESENCE_RUN_ARGS` ?

thomas-riccardi: they are arguments for `telepresence`, not for `telepresence .*run`.
if it's to signify it's used when dmake calls docker run, it could be `DMAKE_RUN_TELEPRESENCE_ARGS`, but it would conflict with `dmake run` which is different than dmake using `docker run`...
=> I'll keep `DMAKE_TELEPRESENCE_ARGS` for now.
It's a non-documented, alpha test, it may be broken in the future.

PR #390, opened 2019-06-10
Dev multi node commands

ibreschi:
To improve the code generation flexibility in dmake, we decided to switch to a node based system.

This big PR resolves multiple problems and should be splitted as following: 
PR1 : Lock bug fix corresponding more or less at the commit 98d97c8 . 
PR2 : Adding the new node based system
PR3 : Integrating the new system in the core.py code generation
PR4 : Removing duplicate code in common.py  TBD
PRN : Others small refactoring like cheking if stages name appear twice


ibreschi:
Closed for inactivity

PR #398, opened 2019-08-12
Remove docker_opt is_public clean

### LINKED ISSUE #397: Remove automatic docker_opts clean
To run the Myriad tests, we need to run docker with the privilege flag.
Unfortunately dmake removes the --privileged docker_opts when the build is not local. 
We agreed with @thomas-riccardi that this protection is not very useful so we will remove it.

ibreschi:
Closes #397 

We need to access `/dev` with `--privileged` on Jenkins for hardware support.
Unfortunately dmake removed `--privileged` from `docker_opts` when the build was not local.
It was not really useful a protection: just remove it.

### DIFF
--- a/dmake/deepobuild.py
+++ b/dmake/deepobuild.py
@@ -890,8 +890,7 @@ class DeployConfigSerializer(YAML2PipelineSerializer):
             opts.append("-v %s:%s" % (common.join_without_slash(host_volume), common.join_without_slash(volume.container_volume)))
 
         docker_opts = self.docker_opts
-        opts = docker_opts + " " + (" ".join(opts))
-        return opts
+        return docker_opts + " " + (" ".join(opts))
 
 class DeploySerializer(YAML2PipelineSerializer):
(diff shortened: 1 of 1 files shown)


thomas-riccardi:
Please merge with useful description in commit.

dmake/deepobuild.py
```diff
@@ -890,10 +890,7 @@ def full_docker_opts(self, env, mount_host_volumes, use_host_ports=None):
             opts.append("-v %s:%s" % (common.join_without_slash(host_volume), common.join_without_slash(volume.container_volume)))
 
         docker_opts = self.docker_opts
-        if not common.is_local:
-            docker_opts = docker_opts.replace('--privileged', '')
-        opts = docker_opts + " " + (" ".join(opts))
-        return opts
+        return docker_opts + " " + (" ".join(opts))
```
thomas-riccardi: I was talking about the line 892: `docker_opts = self.docker_opts`; the `return opts` could be kept

PR #400, opened 2019-08-29
Always chmod 777 the shared volumes

thomas-riccardi:
Previously we did it for dmake shell so there is no issue related to
DMAKE_UID, but in fact we also need it for dmake run/test: sometimes
the containers have non root users and must be able to write to the
shared volume anyway.

Furthermore, the kubernetes equivalent (emptyDir) does that too.


Zanoellia:
Sounds good to me :white_check_mark: I let @maingoh do the approval

PR #402, opened 2019-09-04
Force read and write YAML in version 1.1

### LINKED ISSUE #401: dmake kubernetes deployment broken with container env value 'no'
kubernetes manifest template:
```
apiVersion: apps/v1
kind: Deployment
...
spec:
...
    spec:
      containers:
      - name: foo
        env:
        - name: FOO
          value: 'no'
```
dmake injects extra annotations and renders:
 ```
apiVersion: apps/v1
kind: Deployment
...
spec:
...
    spec:
      containers:
      - name: foo
        env:
        - name: FOO
          value: no
```

diff:
`'no'` became `no`, which in YAML 1.1 is a boolean, not a string. (In YAML 1.2 it is a string.)

This then breaks when applying on kubernetes: the env value must be a string.

### PR DESCRIPTION
**Pull request by thomas-riccardi**
`ruamel.yaml` uses version 1.2 by default, and emits version directive
when forced in 1.1 (`%YAML 1.1` header).

`kubectl` doesn't support explicit `%YAML 1.1` header, nor explicit or
even implicit 1.2, at least regarding the boolean literals issue, see
https://github.com/kubernetes/kubernetes/issues/34146

We now force 1.1 on `ruamel.yaml` and disable emitting the `%YAML 1.1`
version directive.

This fixes `'no'` as container env value in the kubernetes manifests.

Closes #401

### DIFF
--- a/dmake/common.py
+++ b/dmake/common.py
@@ -5,6 +5,7 @@ import logging
 import subprocess
 import re
 from ruamel.yaml import YAML
+from ruamel.yaml.emitter import Emitter as YAML_Emitter
 import uuid
 
 # Set logger
@@ -70,9 +71,16 @@ class FlagBooleanAction(argparse.Action):
 
 ###############################################################################
 
+class YAMLEmitterNoVersionDirective(YAML_Emitter):
+    def write_version_directive(self, version_text):
+        # disable emitting version directive (%YAML 1.1)
+        pass
+
 def yaml_ordered_load(stream, all=False):
     try:
         yaml = YAML(typ='safe', pure=True)
+        # kubectl and everyone else uses yaml 1.1
+        yaml.version = '1.1'
         data = list(yaml.load_all(stream)) if all else yaml.load(stream)
         return data
     except Exception as e:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by thomas-riccardi** [comment]
See also https://bitbucket.org/ruamel/yaml/issues/298/optionally-not-print-version-directive

File: dmake/common.py [hunk 1]
```diff
@@ -70,9 +71,16 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+class YAMLEmitterNoVersionDirective(YAML_Emitter):
+    def write_version_directive(self, version_text):
+        # disable emitting version directive (%YAML 1.1)
+        pass
+
 def yaml_ordered_load(stream, all=False):
     try:
         yaml = YAML(typ='safe', pure=True)
+        # kubectl and everyone else uses yaml 1.1
+        yaml.version = '1.1'
```
maingoh: Maybe set a variable `DEFAULT_YAML_VERSION = 1.1` ?

Zanoellia:
Good for me too :heavy_check_mark:

PR #407, opened 2019-09-17
Support extra characters in branch name

### LINKED ISSUE #23: Dmake fail when there is a `#` or `/` in the branch name
Docker gives a ``` invalid reference format ``` when launching Dmake in a branch with a ```#``` in the name.

thomas-riccardi:
Closes #23, dmake should finally support dependabot.

- Add docker mode for sanitation, and add a unique variant
- Factorize docker image name prefix in a new global: name_prefix
- Sanitize the default docker image tag name: git branches can contain invalid characters
- Support git branches with slashes

### DIFF
--- a/dmake/docker_image.py
+++ b/dmake/docker_image.py
@@ -93,7 +93,7 @@ class ServiceDockerCommonSerializer(YAML2PipelineSerializer, AbstractDockerImage
     name             = FieldSerializer("string", optional = True, help_text = "Name of the docker image to build. By default it will be {:app_name}-{:service_name}. If there is no docker user, it won be pushed to the registry. You can use environment variables.")
     base_image_variant = FieldSerializer(["string", "array"], optional = True, child = "string", help_text = "Specify which `base_image` variants are used as `base_image` for this service. Array: multi-variant service. Default: first 'docker.base_image'.")
     check_private    = FieldSerializer("bool", default = True, help_text = "Check that the docker repository is private before pushing the image.")
-    tag              = FieldSerializer("string", optional = True, help_text = "Tag of the docker image to build. By default it will be '[{:variant}-]{:branch_name}-{:build_id}'")
+    tag              = FieldSerializer("string", optional = True, help_text = "Tag of the docker image to build. By default it will be '[{:variant}-]{:branch_name}-{:build_id}', sanitized and made unique with a hash suffix if needed")
 
     def get_image_name(self, env=None, latest=False):
         if env is None:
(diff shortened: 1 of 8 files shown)


dmake/docker_image.py
```diff
@@ -105,7 +105,7 @@ def get_image_name(self, env=None, latest=False):
             name = common.eval_str_in_env(self.name, env)
         # tag
         if self.tag is None:
-            tag = common.branch.lower()
+            tag = common.sanitize_name_unique(common.branch, mode='docker')
```
maingoh: it was lowered before and not now, is it ok ?

dmake/test_common.py
```diff
@@ -11,10 +11,32 @@
...
+    ('foo#bar', 'foo-bar'),
+    ('foo##bar', 'foo-bar'),
+    ('foo/bar#baz', 'foo-bar-baz'),
+])
+def test_sanitize_name_docker(test_input, expected):
+    assert sanitize_name(test_input, mode='docker') == expected
+    assert re.match(r'[a-zA-Z0-9][a-zA-Z0-9_.-]+', sanitize_name(test_input, mode='docker'))
+
+def test_sanitize_name_unique():
+    assert sanitize_name_unique('foo_bar', mode='docker') == 'foo_bar', "When no sanitation is needed it should return identity"
```
maingoh: What does pytest return if it fails ? Only your error message or also the difference between values ? 
My point is, is it important to add a custom message in the assert from the tests ?

dmake/utils/dmake_run_docker
```diff
@@ -29,11 +29,7 @@ NAME=$2
 shift 2
 
 if [ -z "${NAME}" ]; then
-    if [ ! -z "${BUILD}" ]; then
```
maingoh: It seems BUILD could be empty, if so we might have a double `.` in the base name ?

dmake/docker_image.py
thomas-riccardi: yes, it should never have been: it creates collisions: it could delete official images; i'll precise this in the commit message.

dmake/test_common.py
thomas-riccardi: It serves as documentation for the assert; it is displayed by pytest:
```
    def test_sanitize_name_unique():
>       assert sanitize_name_unique('foo_bar', mode='docker') == 'foo_bar2', "When no sanitation is needed it should return identity"
E       AssertionError: When no sanitation is needed it should return identity                                                                                                                                                               
E       assert 'foo_bar' == 'foo_bar2'                                                                                                                                                                                                       
E         - foo_bar                                                                                                                                                                                                                          
E         + foo_bar2                                                                                                                                                                                                                         
E         ?        +                                                                                                                                                                                                                         
                                                                                                                                                                                                                                             
dmake/test_common.py:40: AssertionError     
```
These are non trivial asserts, I thought the doc would help

dmake/utils/dmake_run_docker
thomas-riccardi: it's indeed a change, but:
1/ it's OK: double dots are allowed by docker
2/ it never happens: on jenkins the build id is always there, and locally it's always `0`

PR #409, opened 2019-10-01
Better tmp dir/files names for easier debug

thomas-riccardi:


dmake/utils/dmake_make_tmp_dir
```diff
@@ -10,9 +10,11 @@ test "${DMAKE_DEBUG}" = "1" && set -x
 
 set -e
 
+NAME=$1
```
thomas-riccardi: make it mandatory

dmake/common.py
```diff
@@ -482,6 +480,11 @@ def init(_options, early_exit=False):
     # Generate name prefix: readable, unique, stable identifier
     name_prefix = sanitize_name_unique('{repo}.{branch}.{build_id}'.format(repo=repo, branch=branch, build_id=build_id), mode='docker')
 
+    if 'DMAKE_TMP_DIR' in os.environ:
```
maingoh: nit: might be `os.environ.pop('DMAKE_TMP_DIR', None)` (avoid copying key)

maingoh: actually, why do you need to remove it, if you overwrite it after ?

PR #417, opened 2019-12-09
Explicitely trying to find python3 first then fallback to python2.

maingoh:
`python` command may not exist everywhere


thomas-riccardi:
python2 will be deprecated in less than one month, let's not add more legacy to dmake.

PR #419, opened 2020-01-14
Add DMake internal plan graph tests

thomas-riccardi:
Test several dmake commands, and compare expected internal graph (via
the `--debug-graph` feature).

DMake generates an internal graph planning the various steps to
execute, according to the command-line and the dmake.yml files.

Example:
service foo with needed_service bar, `dmake test foo` execution plan
graph would have the following edges:
- test foo -> build foo
- test foo -> run bar
- build foo -> build-base foo
- run bar -> test bar
- run bar -> build bar
- test bar -> build bar
- build bar -> build-base bar

This graph can be explored with `dmake --debug-graph` options.

To make execution plan construction easier when adding features to
DMake, this PR adds tests comparing expected with actual execution
plan graph dumps.

The reference graphs can be generated with:
```
make -C test rebuild_graphs
```

Required some cleanup:
- split `argparser` construction out of `dmake` and into `cli.py` to
  be importable
- introduce deepbuild.reset() to reset static classes
  states (just SharedVolumes for now)
- stabilize graph dump order: both ordered dump and ordered
  NeededServiceSerializer env* str/repr.

Also:
- less sys.exit()
- more verbose pytest output to show full diffs on error
- fix/improve test/ project, notably add 2 needed_services on same
  service, but different specialization, to test an edge case on
  NeededServiceSerializer sort.

### DIFF
--- a/dmake/core.py
+++ b/dmake/core.py
@@ -722,7 +722,7 @@ def make(options, parse_files_only=False):
     if auto_complete and auto_completed_app is None:
         raise DMakeException("Could not find any app or sub-app matching '%s'" % app)
 
-    # Filter base images which are not provided
+    # Remove base images which are not provided (by dmake.yml definitions): they are external base images
     for deps in service_dependencies.values():
         to_delete = []
         for i, dep in enumerate(deps):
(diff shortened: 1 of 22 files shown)


thomas-riccardi:
Marking `PAUSED` for now: we have order differences between python 3.5 and 3.6 in the dot output.
We first have order diffs on dump input: `dependencies` (and `attributes`). Possibly different `dict` implem?
Even when forcing the order when generating the dot graph, dot.source re-differs between python 3.5 and 3.6: we have no control of dot serialization order (they probably also use `dict` internally).

Solutions:
- either build a dot graph comparator
- or just compare the dependencies list (force sorted)
  sort `dependencies`, sort `attributes`, still need to also sort env in NeededServiceSerializer 
  or use ordered dict everywhere...

thomas-riccardi:
For reference:
To regenerate the full expected graphs:
```shell
function generate() {
  command=$1
  service=$2
  dmake --debug-graph-and-exit --debug-graph-output-filename "test/test-resources/graph.${command}.${service}.gv" ${command} "${service}"
}
cd test/
generate shell test-web
generate build test-web
generate test test-e2e
generate test test-web2
generate test '*'
generate run test-web2
generate run test-e2e
generate deploy test-k8s
generate deploy test-e2e
generate deploy '*'
```

dmake/core.py
```diff
@@ -757,10 +757,10 @@ def make(options, parse_files_only=False):
     sorted_leaves = filter(lambda a_b__c: a_b__c[0][0] == common.command, sorted_leaves)
     build_files_order = order_dependencies(service_dependencies, sorted_leaves)
 
-    common.dump_dot_graph(service_dependencies, build_files_order)
+    debug_dot_graph = common.dump_dot_graph(service_dependencies, build_files_order)
```
maingoh: this function is hard to review, it should be splitted

test/test_graph.py
```diff
@@ -0,0 +1,38 @@
...
+
+    common.sub_dir = 'test'  # to test '*' for only 'test/'
+    common.generate_dot_graph = True
+    common.exit_after_generate_dot_graph = True
+    common.dot_graph_filename = None
+    dot = core.make(args)
+
+    # generate expected result from `test/` dir with:
+    # `dmake --debug-graph-and-exit --debug-graph-output-filename "test/test-resources/graph.${command}.${service}.gv" ${command} "${service}"`
+    expected_dot_path = 'test-resources/graph.{command}.{service}.gv'.format(command=command, service=service)
```
maingoh: everytime the dmake structure change you will have to change the graph test-resources ?

Issue #223, opened 2018-05-29
More tests types: source tests (style check, static analysis..), unit tests vs integration tests


thomas-riccardi:
Currently dmake executes tests on the built docker image, potentially relying on the whole service dependency tree: this looks like almost integration tests (not really that because the service we test is not started, we just use its docker image).

I see different tests types:
- source tests: style check, static analysis: cannot always be done in the built docker image: we may have removed the sources at that point
- unit tests: maybe without dependencies? mock every dependency? what about db? maybe only allow external dependencies (docker_link). Currently we need to keep the unit tests built binaries in the final service docker image to be able to execute them, it's not great.
- integration/end-to-end tests: the service is really started with the built docker image and all its services dependencies, and we test its public behavior. Currently we have to manually start the service alongside the tests to execute, and the tests are executed from the same docker image *and* container. We could have a different container for these tests. This is currently doable by creating a separate dmake service for each service we wish to test that way, this is not great.

thomas-riccardi:
# Source tests 
They could work like dmake shell (maybe with read-only source volume mount)
This way we can access git submodules (which technically points to a directory in the root `.git/`)

thomas-riccardi:
# Unit testing
Executing tests without running dependencies is now doable with #421 .

PR #421, opened 2020-01-14
Allow faster unit-tests without dependencies at test-time

thomas-riccardi:
related on #223

Fine-grained control of needed_services usage: for parent run, for parent test.

# Allow faster unit-tests without dependencies at test-time.

Later: For fully independent unit-tests we would need to also control
needed_links; possibly by merging needed_links into needed_services.
In the meantime it's fine as needed_links do not recursively depend on
anything.

When working on a specific service (or on services found via change
detection), the `test` step previously always started (recursively)
the needed_services (and also started the links services): dmake
assumed they were integration tests.

Now we can individually disable starting each needed_services: for
run, and test.

Example:
```yaml
services:
- service_name: foo
  needed_services:
  - service_name: bar
    needed_for:
      run: true   # foo needs bar to run
      test: true  # foo needs bar to execute its tests
```

Tests:
- `dmake test test-web2` should not start test-worker anymore (but
  still start rabbitmq)
- `dmake run test-web2` should still start test-worker and rabbitmq
- `dmake test test-e2e` should start rabbitmq and test-worker, but
  test-worker should not be started and linked to test `test-web2`
  itself anymore.

# Also
Internal graph dependency bugfix: we didn't properly support services with needed links but no needed services.

### DIFF
--- a/dmake/core.py
+++ b/dmake/core.py
@@ -194,10 +194,11 @@ def activate_link(loaded_files, service_providers, service_dependencies, service
 
 ###############################################################################
 
-def activate_needed_services(loaded_files, service_providers, service_dependencies, needs, command):
+def activate_needed_services(loaded_files, service_providers, service_dependencies, needs, command, needed_for):
     children = []
     for service, service_customization in needs:
-        children += activate_service(loaded_files, service_providers, service_dependencies, command, service, service_customization)
+        if service_customization is None or service_customization.needed_for.kind(needed_for):
+            children += activate_service(loaded_files, service_providers, service_dependencies, command, service, service_customization)
     return children
 
 ###############################################################################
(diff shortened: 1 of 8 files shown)


dmake/deepobuild.py
```diff
@@ -1352,7 +1353,7 @@ def generate_build_docker(self, commands, service_name):
         service = self._get_service_(service_name)
         service.config.docker_image.generate_build_docker(commands, self.__path__, self.docker, self.build)
 
-    def _launch_options_(self, commands, service, docker_links, run_base_image, mount_root_dir, force_workdir, additional_env = None, additional_env_variables = None, use_host_ports = None):
+    def _launch_options_(self, commands, service, docker_links, run_base_image, mount_root_dir, force_workdir, additional_env = None, additional_env_variables = None, use_host_ports = None, with_dependencies = True):
```
maingoh: nit: default parameters without space around `=`

dmake/core.py
```diff
@@ -211,10 +221,13 @@ def activate_service(loaded_files, service_providers, service_dependencies, comm
             children += activate_base(base_variant)
         elif command == 'test':
             children += activate_service_shared_volumes(loaded_files, service_providers, service)
-            if common.options.with_dependencies and needs is not None:
+            test_with_dependencies = False
+            if common.options.with_dependencies:
```
maingoh: what is the difference between `common.options.with_dependencies` and `test_with_dependencies ` ?

dmake/deepobuild.py
thomas-riccardi: yes but I prefer to follow local standard than mix things; until we clean everything and enforce it.

dmake/core.py
thomas-riccardi: `common.options.with_dependencies` is the `--standalone,-s` vs `--dependencies,-d`; it's a hack to cut all local dependencies and essentially use remote environment.

`test_with_dependencies` is the new feature this PR introduces: some services declare they don't need runtime dependencies to execute their tests, essentially unit-tests.

Here we combine both values to build the dependency tree.

thomas-riccardi:
/rebuild with more data 
and new lines !

dmake/deepobuild.py
```diff
@@ -1123,6 +1133,11 @@ def create_variant(self, variant):
 
         return service
 
+    def get_needed_services_for(self, kind):
+        return [needed_service
```
maingoh: Would be nice to avoid complex list comprehension

Issue #424, opened 2020-02-07
Test parent dependencies in change detection mode


thomas-riccardi:
## Context
Change detection was (re) added by #420.
This re-triggered the question of parent testing (to validate API changes): #55.

Let's re-formulate and better specify #55.

## Goal
To test API changes we rely on executing tests of parent dependencies.

- Before #420 we run `dmake test *`, so we test everything, so all is good.
- With #420 we only test services which have changed. To continue testing API changes we also want to test the parent services.

## Spec questions
1. do we recursively test parents of services which have changed? or just direct parents?
2. do we test again down the dependency tree from these triggered tests on parents?
3. do we want to support `dmake deploy +` too? it adds more questions: do we want to trigger parent deploys too? recursively?

PR #428, opened 2020-02-07
Change detection mode: test parent dependencies

thomas-riccardi:
MVP for #424

When in change detection mode, we work on all changed services, and test their parent recursively.

Doesn't change anything for other steps than `test`. Notable `dmake deploy +` will still only deploy changed services, not its tested parents.

Can be overridden with `needed_services[].needed_for.trigger_test` boolean.

## TODO later
- [ ] maybe don't recurse in the parents?
- [ ] probably down recurse down from the triggered parents.

### DIFF
--- a/dmake/core.py
+++ b/dmake/core.py
@@ -114,12 +114,27 @@ def load_dmake_files_list():
 def add_service_provider(service_providers, service, file, needs = None, base_variant = None):
     """'service', 'needs' and 'base_variant' are all service names."""
     common.logger.debug("add_service_provider: service: %s, needs: %s, variant: %s" % (service, needs, base_variant))
+    trigger_test_parents = set()
     if service in service_providers:
-        existing_service_provider, _, _ = service_providers[service]
-        if existing_service_provider != file:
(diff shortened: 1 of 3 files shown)


dmake/core.py
```diff
@@ -114,12 +114,27 @@ def load_dmake_files_list():
...
+    # add `service` to `trigger_test_parents` backlink for each needed_service child which was asked to trigger the parent test
+    # `trigger_test_parents` == reverse link of `needs`, filtered on `needed_for.trigger_test`
+    if needs is not None:
+        for child_service, child_service_customization in needs:
+            if not child_service_customization.needed_for.kind('trigger_test'):
+                continue
+
+            if child_service in service_providers:
+                # TODO do we need to do something for child_service_customization?
+                service_providers[child_service][3].add(service)
```
maingoh: The fact that we need to hardcode `[3]` is weird :D

dmake/core.py
```diff
@@ -114,12 +114,27 @@ def load_dmake_files_list():
...
+            else:
+                service_providers[child_service] = (None, None, None, set([service]))
```
maingoh: Using dict would probably be more readable.

PR #431, opened 2020-02-18
Support environment variables substitution in kubernetes config_map from_files paths

thomas-riccardi:
like it's already supported in secrets.

There is still another difference: we assume it's a relative path for
config_map, and an absolute path in secrets, but it should not change much.


thomas-riccardi:
not needed

Issue #432, opened 2020-02-27
Environment variables values should always be strings in dmake.yml


thomas-riccardi:
```yaml
dmake_version: 0.1
app_name: env-test

env:
  default:
    variables:
      FOO: on

docker:
  root_image:
    name: ubuntu
    tag: 18.04

services:
  - service_name: foo
    config:
      docker_image: ubuntu:18.04
    tests:
      commands:
        - echo $FOO
```

Actual:
```
$ dmake test
...
 ## Running App ##
- Running test @ env-test/foo
+ echo True
True
```

Expected:
```
on
```

This is due to
- yaml having many boolean literals: https://yaml.org/type/bool.html
- dmake accidentally accepting booleans as environment variables values, because:
  -  dmake tries to convert non strings to strings there: https://github.com/Deepomatic/dmake/blob/3248370f76a59373b0c9dcf6c498c090ab134c2c/dmake/serializer.py#L130-L141
  - dmake doesn't explicitly supports booleans, but in python `isinstance(True, int)` returns True.

What I propose is a slow migration toward what kubernetes does: when it wants strings it mandates string type in yaml: early failure is better than unexpected env var values at runtime...

PR #433, opened 2020-02-28
Better yaml values usage: warn about unexpected use-cases

thomas-riccardi:
related to #432

warning for #432 example:
```
[DEPRECATION WARNING] D002: Field 'env:default:variables' in 'dmake.yml' should contain a string, it currently contains a <class 'bool'>: True. Add quotes arount the value.
[DEPRECATION WARNING] D002: Field 'docker:root_image:tag' in 'dmake.yml' should contain a string, it currently contains a <class 'float'>: 18.04. Add quotes arount the value.
[DEPRECATION WARNING] D004: Field 'services:config:ports:container_port' in 'dmake.yml' should contain an int, it currently contains a bool: True.
```

### DIFF
--- a/dmake/serializer.py
+++ b/dmake/serializer.py
@@ -1,6 +1,7 @@
 import os, sys
 import copy
 from collections import OrderedDict
+from numbers import Number
 from dmake.common import DMakeException
 import dmake.common as common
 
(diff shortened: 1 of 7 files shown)


dmake/serializer.py
```diff
@@ -126,9 +151,16 @@ def _validate_type_(self, file, needed_migrations, data_type, data, field_name):
         elif data_type == "int":
             if not isinstance(data, int):
                 raise WrongType("Expecting int")
+            if isinstance(data, bool):
+                common.logger.warning("[DEPRECATION WARNING] D004: Field '{field}' in '{file}' should contain an int, it currently contains a bool: {value}.".format(field=field_name, file=file, value=data))
+            return data
+        elif data_type == "number":
```
maingoh: Might worth considering enums one day

thomas-riccardi: When we drop python 2 yes!

PR #435, opened 2020-03-12
Better docker container names for debug: now contains service name

thomas-riccardi:


dmake/utils/dmake_run_docker_daemon
```diff
@@ -28,6 +28,9 @@ SERVICE_NAME=$1; shift
 LINK_NAME=$1; shift
 NAME=$1; shift
 
+if [ -z "${NAME}" ]; then
+  NAME_PREFIX=$"${NAME_PREFIX}.$(echo ${SERVICE_NAME} | sed 's/[^a-zA-Z0-9_.-]/_/g')"
```
maingoh: is the $ before the `"` on purpose ?

thomas-riccardi: good catch !

Issue #415, opened 2019-11-27
Validate kubernetes deployment (dry-run) from PR builds in Jenkins


thomas-riccardi:
When testing a PR on jenkins we only execute `dmake test`, and the branch name is usually *not* a deployed branch. 2 reasons the deployment part is not tested.
(This leads to deployment configuration error being detected too late: after the PR is merged...)

(The usual errors are: kubernetes dmake templates escaping error: `$` non escaped, invalid kubernetes manifest, both are testable at dmake generation time). 

## Manually
`DMAKE_ON_BUILD_SERVER=1 dmake deploy '*' -b master` works great to test deployment.

## On Jenkins
- maybe add an extra `dmake deploy` fake? or remplace `dmake test` by `dmake deploy --deploy-dry-run` ?
- somehow do the `dmake deploy` dry-run part on target branch instead of source branch. It cannot be done simply by replacing `branch` by the target branch as the per-PR-uniqueness is used to generate unique resource names.

thomas-riccardi:
`utils/dmake_deploy_kubernetes` already supports `DMAKE_K8S_DRY_RUN=1` env var, but we probably also want to skip docker push (and there is still the branch issue).

thomas-riccardi:
Starting with k8s 1.18 there is also the notion of server dry-run: https://kubernetes.io/docs/reference/using-api/api-concepts/#dry-run

thomas-riccardi:
Just running the planning/generation phase alas does not catch wrong yaml type in `env[].value` (but does catches unknown fields...).

We could apply that with server dry-run when server are at 1.18 version

thomas-riccardi:
#449 is a good enough solution for now

PR #449, opened 2020-05-05
Test kubernetes deployment manifest templates in PRs

thomas-riccardi:
MVP for #415

It works by just planning/generating the deployment step (in practice
also the build step as it's not skippable...): all files are read,
template instantiated, and somewhat validated by kubectl dry-run.

It thus catches:
- missing referenced files
- wrong templates (undefined variable)
- some wrong k8s manifest: unknown fields, etc, but not all
  issues (e.g. env[].value yaml type is *not* checked to be a string)
  - it does check some schema validation via remote api definition (properly since #358)

It also supports no deployment (i.e. no `deploy` section (which is different than empty `deploy` array, which asks dmake to push docker images), now required when images are local only (no `foobar/` prefix)).

This will still test the k8s deploy dry-run with `DMAKE_SKIP_TESTS=1`, which is useful. To really skip that: change the DMAKE_COMMAND: not empty nor test (which means dry-run test deployment is mandatory when asking for tests, which is fine).

Later it could be integrated to dmake itself, but there are tricky
parts:
- is it a new command? or included in `test` ? or in `deploy`?
- we do need the target branch to trigger the deployment (there is a
  branch whitelist in dmake.yml controlling deployment)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,6 +6,9 @@ properties([
         string(name: 'BRANCH_TO_TEST',
                defaultValue: env.CHANGE_BRANCH ?: env.BRANCH_NAME,
                description: 'The branch to check out. Only used when testing a directory different from deepomatic/dmake.'),
+        string(name: 'DEPLOY_BRANCH_TO_TEST',
+               defaultValue: env.CHANGE_TARGET ?: 'master',
+               description: 'The target branch to use for kubernetes deployment dry-run test.'),
         string(name: 'DMAKE_APP_TO_TEST',
                defaultValue: '*',
                description: 'Application to test. You can also specify a service name if there is no ambiguity. Use * to force the test of all applications.'),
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -109,6 +119,13 @@ node {
         sh ". .venv3/bin/activate && pytest -vv --junit-xml=junit.xml --junit-prefix=python3"
         junit keepLongStdio: true, testResults: 'junit.xml'
       }
+      if (params.DMAKE_COMMAND == 'test') {
```
maingoh, line 122: this look easily factorizable with the python2 version

dmake/templates/jenkins/Jenkinsfile
```diff
@@ -128,6 +128,15 @@ sshagent (credentials: (env.DMAKE_JENKINS_SSH_AGENT_CREDENTIALS ?
 
     dmake_with_dependencies = params.DMAKE_WITH_DEPENDENCIES ? '' : '--standalone'
 
+
+    if (is_pr && dmake_command == 'test') {
+        echo "First: kubernetes deploy dry-run (just plan deployment on target branch to validate kubernetes manifests templates)"
```
maingoh, line 133: what about having a function like:
```
def dry_run_deploy(cmd_prefix, dmake_exe, app, branch):
    pass

dry_run_deploy('', '\$(which dmake)', '${params.DMAKE_APP}', '${env.CHANGE_TARGET}')
dry_run_deploy('. .venv2/bin/activate && ', 'dmake'', '${params.DMAKE_APP_TO_TEST}', '${env.DEPLOY_BRANCH_TO_TEST}')
dry_run_deploy('. .venv3/bin/activate && ', 'dmake', '${params.DMAKE_APP_TO_TEST}', '${env.DEPLOY_BRANCH_TO_TEST}')
```

PR #453, opened 2020-05-27
Drop python2 support

### LINKED ISSUE #434: Drop python 2 support
Python 2 support currently holds us back on some features:
- #229
- we have (had?) some instability in dict order in python2
- missing enums
- ...

We have more complex code to support python 2.

And python 2 is sunset: https://www.python.org/doc/sunset-python-2/

=> Let's prepare python 2 support sunset in dmake.

Dependency:
- full dmake execution without python2: #161
- once read, maybe start by warning about python 2?

thomas-riccardi:
Closes #434

Also:
- supports ubuntu 20.04 with python 3.8
- bump dependencies notably to support python 3.8
  - notably update `kubernetes` python client which should better support new cluster versions and some interesting bugfixes around auth with gcp and azure.


dmake/docker_config.py
```diff
@@ -73,19 +74,18 @@ def docker_credentials_store(store, command, input):
...
     """
     credentials_store, registry, data = get_docker_config_auth(registry_url)
     if credentials_store is None:
         # get from config file data
-        # TODO parse value and return a (user,password) tuple instead of a requests kwargs
-        headers = {'Authorization': 'Basic %s' % data['auth']}
-        return {'headers': headers}
+        auth = data['auth']
+        username_password = base64.b64decode(auth).decode("UTF-8")
+        return tuple(username_password.split(':'))
```
putze, line 87: This is causing a problem if the password contains a `:`

PR #463, opened 2020-08-14
Release command: fix prerelease bumping

### LINKED ISSUE #462: Release command: fail when bumping pre-release part
## Description

There is an error when bumping the pre-release part of a version number already including a pre-release part, thus making release impossible.

## How to reproduce

```bash
# deepomatic-rpc, master branch
# current version is v0.8.3-dev2

$ version=$(bump2version --dry-run --list 0 | grep current_version | sed -r s,"^.*=",,)
$ echo $version
v0.8.3-dev3

$ dmake release -t "v${version}" deepomatic-rpc
ERROR: Could not find any corresponding correct candidate previous version when bumping to v0.8.3-dev3. Previous version candidate: v0.8.3-dev2
```

Zanoellia:
Close #462 (fix prerelease bumping)

- **Order** The check ensure the new version is strictly superior to the previous version.
- **Main parts** The check still ensure no version is skipped when bumping the major/minor/patch part or finalizing (=removing extensions, i.e. prerelease and build parts).
- **Prerelease** The releaser has now more flexibility on the pre-releases: they are not checked too much because they can vary a lot. For instance token can change between version (`alpha` -> `beta`) or between projects (`rc`, `dev`...) (https://semver.org/#spec-item-9). Only the alphabetical order is checked if all the other parts are identical
- **Build** If only the build part differs, there is no precedence (https://semver.org/#spec-item-10) and the bump is invalid (arbitrary choice, can change later)

---

Additional changes:
- Light refactoring
- Add a test on the version checking method

### DIFF
--- a/dmake/commands/release.py
+++ b/dmake/commands/release.py
@@ -26,11 +26,27 @@ def tag_to_version(tag):
 
 
 def is_valid_bump(prev_version, next_version):
+    """Returns True if the bump is valid according to Semantic Versionning
+
+    A bump is valid if:
+    -   The release version is strictly higher than the current version, including extensions
+    -   No version is skipped when bumping the major/minor/patch part or
+        finalizing (=removing extensions, i.e. prerelease and build parts).
+
+    Note that a bump is invalid if only the build part differs. This behaviour may
+    be too restrictive and change later.
+
+    More info: https://semver.org or see examples in test/test_release_command.py
+
+    Args:
+        prev_version (VersionInfo): the current version
+        next_version (VersionInfo): the version to release
+    """
     if prev_version >= next_version:
         return False
 
-    # Additional check: ensure no version is skip
-    # Pre releases and builds are not taken in account as their token can change
+    # Ensure no version is skipped
+    # Prereleases and builds are not taken in account as their token can change
     # Example: (1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-beta)
     if prev_version.prerelease is None and next_version.prerelease is None and \
             prev_version.build is None and next_version.build is None:
@@ -38,6 +54,7 @@ def is_valid_bump(prev_version, next_version):
                semver.bump_minor(str(prev_version)) == str(next_version) or \
                semver.bump_patch(str(prev_version)) == str(next_version)
 
+    # Ensure no version is skipped if finalizing the version
     elif prev_version.prerelease is not None and next_version.prerelease is None:
         return tag_to_version(semver.finalize_version(str(prev_version))) == next_version
     else:
--- a/dmake/commands/release.py
+++ b/dmake/commands/release.py
@@ -1,10 +1,10 @@
 import os
+
 import dmake.common as common
 import inquirer
 import semver
-from github import Github
-
 from dmake.common import DMakeException
+from github import Github
 
 
 def remove_tag_prefix(tag):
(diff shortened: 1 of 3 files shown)


dmake/commands/release.py
```diff
@@ -1,58 +1,87 @@
...
     try:
-        return semver.parse_version_info(remove_tag_prefix(tag))
+        return semver.VersionInfo.parse(remove_tag_prefix(tag))
     except ValueError:
         return None
 
 
+def is_valid_bump(prev_version, next_version):
+    if prev_version >= next_version:
+        return False
```
Zanoellia, line 30: Todo: add comment

test/test_release_command.py
```diff
@@ -0,0 +1,23 @@
...
+    ('0.2.0', '0.3.0', True),
+    ('0.2.0', '1.2.0', False),
+    ('0.2.0', '0.2.2', False),
+    ('0.8.2', '0.8.3-dev1', True),
+    ('0.8.3-dev1', '0.8.3-dev2', True),
+    ('0.8.3', '0.8.3-dev1', False),
+    ('1.0.0-alpha', '1.0.0-alpha.1', True),
+    ('1.0.0-alpha.1', '1.0.0-beta', True),
+    ('1.0.0-beta+0', '1.0.0-alpha.1+45', False)
+])
```
Zanoellia, line 19: TODO: add comments/group examples by type

test/test_release_command.py
```diff
@@ -0,0 +1,23 @@
...
+
+from dmake.commands.release import is_valid_bump
+
+
+@pytest.mark.parametrize("prev_version_str,next_version_str,is_valid", [
+    ('0.2.0', '0.3.0-rc1', True),
+    ('0.2.0', '0.2.1', True),
+    ('0.2.0', '0.3.0', True),
+    ('0.2.0', '1.2.0', False),
+    ('0.2.0', '0.2.2', False),
```
maingoh, line 12: Maybe we want to allow this case ? Do we always want the semver to be continuous ?

maingoh: Also maybe add extra tests like this:
```
    ('0.2.2', '1.0.0', True),
    ('0.2.2-dev1', '1.0.0', True),
    ('0.2.2', '1.0.0-dev1', True),
```

dmake/commands/release.py
```diff
@@ -1,58 +1,87 @@
...
+            latest_per_major_minor[(version.major, version.minor)] = version
 
     # Ask for previous version
     if release_tag is None:
         questions = [
             inquirer.List(
                 'release_tag',
                 message="Here are only the latest tags per major-minor version. Which tag do you want to release on?",
-                choices=[tags_list[key].name for key in sorted_release_keys if (key.major, key.minor) in latest_per_major_minor and latest_per_major_minor[(key.major, key.minor)] == key] + ['Other'],
+                choices=[github_tag_list[key].name for key in sorted_release_versions if
```
maingoh, line 82: Even if you just put it on multiline, this kind of list comprehension is not so much readable to me, I tend to prefer a more verbose `for`.
```python
choices = []
for key in sorted_release_versions:
    major_minor = (key.major, key.minor)
    if major_minor in latest_per_major_minor and latest_per_major_minor[major_minor] == key:
        choices.append(github_tag_list[key].name)
choices.append('Other')
```

test/test_release_command.py
Zanoellia: > Maybe we want to allow this case ? Do we always want the semver to be continuous ?

I don't know, I just kept this existing check. I guess it's more strict and professional to avoid skipping version but we can change it later if we the need arises

dmake/commands/release.py
Zanoellia: To be honest I didn't even looked at it because my eyes cried :joy_cat: , my IDE did everything. Good idea to make it readable and more maintainable.

requirements.txt
```diff
@@ -7,7 +7,7 @@ oauthlib>=2.0.6
 requests-oauthlib==0.8.0
 inquirer==2.2.0
 PyGithub==1.39
-semver==2.8.0
+semver==2.10.2
```
thomas-riccardi, line 10: let's try to not bump python dependencies if possible

dmake/commands/release.py
Zanoellia: Done

test/test_release_command.py
Zanoellia: Done

dmake/commands/release.py
Zanoellia: Done

requirements.txt
Zanoellia: Rollbacked. It's hard for the eyes :joy: 

```python
return is_valid_bump(tag_to_version(semver.finalize_version(str(prev_version))), next_version)
```

I might have a bug with this version though when all the parts are equals except the build parts :thinking: ...

Zanoellia: In fact the bug was in my code :joy_cat:

Zanoellia:
OK I made the rollback, added a few use cases in the tests and comments/docstring in both code and tests.

I let you a few hours before merging in case you have additional feedback

PR #466, opened 2020-10-12
Raise error on empty DMAKE_GITHUB_TOKEN/DMAKE_GITHUB_OWNER env vars

### LINKED ISSUE #459: Exception if DMAKE_GITHUB not set never triggered
## Problem 

I did not setup my `DMAKE_GITHUB` variables and I had the following error:
```bash
~$ dmake release
# [...]
Traceback (most recent call last):
  File "/home/juliette/deepomatic/dmake/dmake/commands/release.py", line 34, in entry_point
    owner = g.get_user(owner)
github.GithubException.BadCredentialsException: 401 {u'documentation_url': u'https://developer.github.com/v3', u'message': u'Bad credentials'}
Makefile:22: recipe for target 'release' failed
make: *** [release] Error 1
```
instead of triggering our proper exceptions defined here:
https://github.com/Deepomatic/dmake/blob/cc81092f08f3de6fa16c5587e4fcdc3925c077fc/dmake/commands/release.py#L25-L30

It is caused by the set of those env variables to `'' in the installation script if not already set by the host's user.

https://github.com/Deepomatic/dmake/blob/cc81092f08f3de6fa16c5587e4fcdc3925c077fc/install.sh#L128-L129

Thus, `os.getenv` will return a empty string instead of None and the proper exception is never triggered

## Reproduction

Set your env variables to `''` then run `dmake release`

## Solution

Fix installation script by (either):
- Do no set up env variables if empty
- Ask the user to provide the missing credentials

thomas-riccardi:
They are set to empty string by the install.sh script...

Closes #459

### DIFF
--- a/dmake/commands/release.py
+++ b/dmake/commands/release.py
@@ -67,10 +67,10 @@ def entry_point(options):
 
     token = os.getenv('DMAKE_GITHUB_TOKEN', None)
     owner = os.getenv('DMAKE_GITHUB_OWNER', None)
-    if token is None:
+    if not token:
         raise DMakeException(
             "Your need to define your Github Access Token by setting the DMAKE_GITHUB_TOKEN environment variable")
-    if owner is None:
+    if not owner:
         raise DMakeException(
             "Your need to define your Github account/organization name by setting the DMAKE_GITHUB_OWNER environment variable")
 


Zanoellia (approved):
:heavy_check_mark:

PR #468, opened 2020-10-21
Add k8s annotations, also in spec.template.metadata to force redeploy with the deploy timestamp

thomas-riccardi:
Before, when running dmake deploy locally, the image name didn't change, so the deployment spec.template usually didn't change (it did change when env var values changed), resulting in new code not deployed until pod deletion.
(in any case this still needs a `imagePullPolicy: Always` for these dev envs so k8s re-downloads the (new) docker image.

Now, set dmake annotations notably with timestamp (so it's unique) not only in metadata  but also in spec.template.metadata when spec.template exists; forcing a new deployment.

Also removed annotations set by dmake_deploy_kubernetes at runtime: it's all done at generation time now.

Later we could also set the kubernetes.io/change-cause this way (but maybe add it only at the top-level? maybe only for app/v1?).

### DIFF
--- a/dmake/common.py
+++ b/dmake/common.py
@@ -1,6 +1,7 @@
 import os
 import sys
 import argparse
+from datetime import datetime
 import hashlib
 import io
 import logging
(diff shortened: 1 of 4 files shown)


Alexis-Jacob (approved):
Working perfectly for me

PR #470, opened 2020-11-05
Change detection trigger test v2: bugfix for non-trivial cases: Don't cascade back from the triggered parents on siblings

thomas-riccardi:
related to #424

### Scenario: `dmake test +`, service Child1 source changes
Issue:
- Child1 test triggered by source change
- Parent test triggered by child source change (technically, when activating Child1 service for command==test)
- Parent test needs Child1 and Child2 run
- Child2 run needs Child2 test <-- issue here: we should not test Child2; but we should keep the "`Child2 run`->`Child2 test`" dependency if for another reason `Child2 test` is created (e.g. Child2 sources have also changed, or Child2 test is triggered by one of its own children source changed (recursively))

### What this PR (last commit) does:
On change detection mode, only create the dependency from run node to
test node if test node already activated otherwise. Previously it
always created the test node in addition to linking run to test to
guarantee execution order.
That previous test node creation effectively cascaded down to most of 
the dependency DAG, resulting in inefficiently testing too much.

Only the services to test by change or parent trigger are explicitly
created, alas it's *not* guaranteed to be done before needing to add the
dependency because we fully work changed service by changed service
sequentially, so when multiple services have changed we don't know yet
if a test will be needed or not.

To work around that we symply add the run->test link ordering in a
second pass in make().

Added assert command != 'run' => service_customization == None in
activate_service as we use that previously implicit property in that
new second pass.

## Various cleanups
To make things easier to understand/debug in this area, this PR starts with multiple cleanup commits, they are all noop (or just change logging):
- Fix error when internal error 'circular dependencies' happens.
- More debug logs: all run_shell_commands
- change detection: add DMAKE_CHANGE_DETECTION_OVERRIDE_DIRS env var for easier tests
- Refactor find_active_files for better clarity
- Refactor: Stop touching `service_dependencies` simply for pruning for debug graph: use dedicated variable
- More aligned activate_service debug logs, with tabs

### DIFF
--- a/dmake/core.py
+++ b/dmake/core.py
@@ -222,6 +222,8 @@ def activate_needed_services(loaded_files, service_providers, service_dependenci
 
 def activate_service(loaded_files, service_providers, service_dependencies, command, service, service_customization=None):
     common.logger.debug("activate_service: command: %s,\tservice: %s,\tservice_customization: %s" % (command, service, service_customization))
+    if command != 'run':
+        assert service_customization == None
     node = (command, service, service_customization)
     if command == 'test' and common.skip_tests:
         return []
@@ -249,8 +251,21 @@ def activate_service(loaded_files, service_providers, service_dependencies, comm
             children += activate_base(base_variant)
         elif command == 'run':
             children += activate_service_shared_volumes(loaded_files, service_providers, service)
+            # ~hackish: run service depends on test service if we are doing tests
             if common.command in ['test', 'deploy']:
-                children += activate_service(loaded_files, service_providers, service_dependencies, 'test', service)
+                if common.change_detection:
+                    # in change detection mode, don't add "test service" node: only create the link between run and test if the test node exists.
+                    # the services to be tested are either:
+                    # - created directly via graph construction starting on the target services: the ones that changed
+                    # - independently created from "test child service" to "test parent service"
+                    # we can reach here in the middle of the DAG construction (e.g. when multiple services have changed: we fully work one by one sequentially),
+                    # so we don't know yet if the test node will exist at the end or not.
+                    # the link will be created later, in a second global pass in make(), see "second pass" there
+                    pass
+                else:
+                    # normal mode, activate "test service" as dependance of "run service"
+                    # REMARK: if we wanted, we could change the semantic of `dmake test foo` to only test foo (while still running its dependencies needed for tests or run, recursively), instead of also testing all children services too: just use the second pass
+                    children += activate_service(loaded_files, service_providers, service_dependencies, 'test', service)
             children += activate_service(loaded_files, service_providers, service_dependencies, 'build_docker', service)
             if common.options.with_dependencies and needs is not None:
                 children += activate_needed_services(loaded_files, service_providers, service_dependencies, needs, command='run', needed_for='run')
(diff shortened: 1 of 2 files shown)


dmake/core.py
```diff
@@ -247,8 +249,21 @@ def activate_service(loaded_files, service_providers, service_dependencies, comm
...
             if common.command in ['test', 'deploy']:
-                children += activate_service(loaded_files, service_providers, service_dependencies, 'test', service)
+                if common.change_detection:
+                    # in change detection mode, don't add "test service" node: only create the link between run and test if the test node exists.
+                    # the services to be tested are either:
+                    # - created directly via graph construction starting on the target services: the ones that changed
+                    # - independently created from "test child service" to "test parent service"
+                    # we can reach here in the middle of the DAG construction (e.g. when multiple services have changed: we fully work one by one sequentially),
+                    # so we don't know yet if the test node will exist at the end or not.
+                    # the link will be created later, in a second global pass in make()
```
vdel, line 261: Maybe be even more explicit in the other code block reference:

```suggestion
                    # the link will be created later, in a second global pass in make() (look for "<second_pass>" below)
```

dmake/core.py
```diff
@@ -817,17 +832,31 @@ def make(options, parse_files_only=False):
         else:
             activate_service(loaded_files, service_providers, service_dependencies, common.command, auto_completed_app)
 
-    # (warninig: tree vocabulary is reversed here: `leaves` are the nodes with no parent dependency, and depth is the number of levels of child dependencies)
+    # second pass
```
vdel, line 835:
```suggestion
    # <second_pass> (see the comment at the other occurrence of <second_pass> above)
```

dmake/core.py
```diff
@@ -817,17 +832,31 @@ def make(options, parse_files_only=False):
...
+    for node in service_dependencies:
+        command, service, service_customization = node
+        if command == 'run' and common.change_detection:
+             # guarantee "always test a service before running it" when change detection mode: run doesn't trigger test, but if test exists for other reasons, we still want to order it after run, see activate_service() command=='run' comments
+            test_node = ('test', service, None)
```
vdel, line 840: If I understand: you assume the form of the test node would be `('test', service, None)` and look if it actually exists. 

Could it be that there is a bug here if your service has some variants ? 

For example (if do not remember the syntax for variants), for worker-nn could it be that `('test', "worker-nn", "caffe")` and `('test', "worker-nn", "tensorflow")` exist but not `('test', "worker-nn", None)`

vdel (approved):
LGTM besides the potential bug.

nit: `change_detection` seems to be False by default ? Why not setting it to True ?

nit: I do understand why the `test node creation effectively cascaded down to most of
the dependency DAG, resulting in inefficiently testing too much.`  ? (I could not find the location in the code where a `test` on a service activates `test` on its dependencies ?)

dmake/core.py
thomas-riccardi: The 3rd part of the node tuple is not the base/service variant, but the runtime needed_services customisation (e.g. I want to run my worker service with a different environment)
Luckily for us, this is non-`None` only for `run` command nodes, so `test` nodes have always `None` there. I'll add a comment in the code indeed.

thomas-riccardi:
@vdel I added an explanation of the issue fixed by this PR at the beginning of the PR description, this was indeed missing.

dmake/core.py
thomas-riccardi: I added an assert in activate_service too.

thomas-riccardi:
@vdel 
> nit: change_detection seems to be False by default ? Why not setting it to True ?

conservative choice, it's enabled via `+` as service name in some projects, for PR only, on jenkins.
Developers can use it locally but it's a little bit harder to setup correctly (it relies on git upstream branch to choose the base for the change comparison; it may be badly setup).

PR #471, opened 2020-12-15
Workaround scripted pipeline 'echo' not always showing up in high-level interfaces

thomas-riccardi:
Some `echo` are displayed in step views, some not, no idea why. it seems the ones talking about `::shared_volume` are shown, others not, no idea why:
before, classic
before, blueocean

May or may not be related to https://issues.jenkins.io/browse/JENKINS-53649

Implemented workaround from https://stackoverflow.com/a/58558537


maingoh (approved):
LGTM !

PR #472, opened 2020-12-18
Parallel Execution on Jenkins

thomas-riccardi:
# Goal: faster CI by executing dmake steps in parallel when possible

This MVP only works on Jenkins, where we have higher-level primitives, and more resources control.

Enable it via `DMAKE_PARALLEL_EXECUTION = 1` environment variable.

# Results
test:`dmake test \*`, with base image and image already in cache, a.k.a. worst case for easy parallel gains (docker builds are all parallelizable)

Full command when debugging this:
```
DMAKE_PARALLEL_EXECUTION=1 CHANGE_BRANCH=master DMAKE_ON_BUILD_SERVER=1 dmake --debug-graph-and-exit --debug-graph-pretty --debug-graph-group-by height deploy \*
```
- `DMAKE_PARALLEL_EXECUTION=1` to enable parallel execution
- `CHANGE_BRANCH=master` to fake Jenkins environment (i.e. a PR to be merged into `master` branch), so it generates with Jenkinsfile instead of bash format (which would fail as parallel is not supported there)
- `DMAKE_ON_BUILD_SERVER=1` to generate the DMakefile and stop there, without executing it
- `--debug-graph-and-exit`: generate debug graph, then exit, skipping plan/DMakefile generation (some of the previous options are thus useless in this narrow case)
- `--debug-graph-pretty` to have pretty visual formatting of the graph (that's new, and that's the default)
- `--debug-graph-group-by height` to generate a debug graph ranked by height: each line is a a parallel execution bloc
- `deploy \*` to have the complete graph with `deploy`

How to read images:
- jenkins blue ocean: displays the sequence of parallel executions from left to right
- dmake debug graph: each node with the real dependencies as arrows, nodes of same height are positioned on the same line: each line is executed in parallel (starting from bottom)

## dmake: 2min15s -> 1min30s
https://build.deepomatic.com/blue/organizations/jenkins/dmake/detail/PR-472/12/pipeline
dmake jenkins blue ocean
dmake dmake-services debug height gv
dmake dmake-services debug height pretty deploy gv

See https://github.com/Deepomatic/squad-infra-deploy/issues/90#issuecomment-756180208 for more information, tests on internal repositories, etc.

# Details
Some preliminary work was done in #474

More preliminary work:
- New height: ASAP vs ALAP; Add --debug-graph-group-by option: optionally sort by height
- Valid generated Jenkinsfiles checks moved into generate_command_pipeline
- Fix uniqueness of NeededService name suffix

Real work:
- Parallel execution by height, with global Jenkins machine resource lock to try to avoid saturating resources (mainly RAM)
- Parallel deploy all services, after all the other nodes (without resource lock to avoid potential kubernetes deployment deadlocks), to avoid half deploying an app if an independent service test fails.

Building blocks:
- use `parallel()` which runs multiple branches of code in parallel,
  in one common executor
  Limitation: need static definition of parallel branches, Jenkins
  doesn't support dynamic generation of branches :-(
- use `lock(PARALLEL_BUILDERS)` to globally limit parallelism on the
  whole Jenkins machine, across jobs, to limit resource
  saturation (mainly RAM)
Locking in a parallel branch doesn't use any (scarce) resource, so the
pattern of `parallel+lock` is OK.

As an MVP, we executes nodes of same height in parallel: the global
execution is a sequence of parallel executions of steps of same
height.
This is not optimal but:
- we cannot reach optimal solution with static scheduling and unknown
  tasks duration,
- it works by construction, nodes depend (directly or indirectly) on
  nodes of strictly lower height, so same height are guaranteed to be
  independent: we can execute them in parallel
- and it gives some notable gains on standard patterns:
  - `base_build` then `build` steps are executed in parallel
  - `needed_links` are started as soon as possible: while the rest
    is built
  - with any luck, some tests are executed in parallel

Later versions could implement some heuristics, maybe detect
independent sub-graphs after having removed base_build and build,
i.e. for run and test, and execute them in parallel (nested
`parallel()`, it works).

From the user point-of-view, we display the standard ordered Plan, by
command, then, as a second pass, display (and dump) the new parallel
plan, by height.

Special case: all `deploy` nodes are executed in parallel, after
everything else (and without resource lock): we don't want to start
deploying some services of an app while other services are still being
tested: their tests could fail and we don't want a partially deployed
app.
Don't lock PARALLEL_BUILDERS on deploy height, it could lead to
deployment deadlock if there is a deployment runtime dependancy
between services.
Caveat: the height on the debug graph is thus wrong in regard to the
parallel execution; it was to much hack to have the graph with
"new/faked" height as in parallel execution while *not* changing
anything to the initial/default Plan.

Implementation-wise, we generate the commands for each node only once:
in the first, default, pass, as before; additionally, we now store
them in a new `nodes_commands[node]` dict.
This is reused in the second pass to re-order by height, and wrap with
`parallel()` and `lock()`.
For the regrouping of `deploy` nodes, we just extract them while
constructing the `nodes_by_height` dict, and inject them back at the
end on a new height.

Special care is taken on GPU lock to start locking it as late as
possible, but since the nodes are now executed as soon as possible, we
may see more GPU contention.
We cannot unlock GPU before the very end of the dmake execution
because there is no notion of `run_stop` for now (we could stop and
unlock gpu before deploying for example)

Visually on Jenkins:

Classic UI:
- logs are prefixed by the parallel branch name
  e.g. `[base @ dmake-test-web-base::base]`
- `Pipeline Steps` show a tree representation, with stage and ~step
  names, durations, and access to log for each step

Blue Ocean:
- the sequence of parallel branches is displayed \o/
- parallel blocs labels are not displayed :-(
- each node is labeled, and clickable to get steps logs
- parallel blocs with zero branches are skipped (otherwise Jenkins
  raises an error)
- parallel blocs with one branch are emitted as such, but create
  graphical glitch on blue ocean with our current jenkins version: the
  node is not visible during the build; but after the job has
  finished, it is visible
  (I tried to emit a stage without `parallel()` in such case, but a
  worse bug happened: the node disappeared not during but after the
  build...)

# Followup
- [ ] use heuristics: parallel `build_base` and `build` is good, but by excluding them independent sub-graphs could be extracted and executed in parallel (nested parallel), effectively working on run+test commands only at that point (it will probably be time to switch to `networkx`); try to optimize for future unit-test+e2e pattern
- [ ] maybe implement for local execution too? using gnu `parallel`? careful with parallelism: resources can be more scarce on developers machines
- [ ] maybe split `deploy` into `publish` and `deploy` (`docker push` and `kubectl apply`): publish can be done with resources lock, only `kubectl apply` needs to be done in parallel to avoid locks. Maybe regroup them into one deploy node per app?

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -87,6 +87,9 @@ node {
     if (params.DMAKE_PAUSE_ON_ERROR_BEFORE_CLEANUP) {
         env.DMAKE_PAUSE_ON_ERROR_BEFORE_CLEANUP=1
     }
+
+    // test parallel execution
+    env.DMAKE_PARALLEL_EXECUTION = 1
   }
   stage('Python 3.x') {
     sh "virtualenv -p python3 workspace/.venv3"
--- a/dmake/cli.py
+++ b/dmake/cli.py
@@ -39,6 +39,7 @@ argparser = argparse.ArgumentParser(prog='dmake')
 argparser.add_argument('--debug-graph', default=False, action='store_true', help="Generate dmake steps DOT graph for debug purposes.")
 argparser.add_argument('--debug-graph-and-exit', default=False, action='store_true', help="Generate dmake steps DOT graph for debug purposes then exit.")
 argparser.add_argument('--debug-graph-group-by', default='command', choices=['command', 'height'], help="Group nodes by <> in generated DOT graph.")
+argparser.add_argument('--debug-graph-pretty', '--no-debug-graph-pretty', default=True, action=common.FlagBooleanAction, help="Pretty or raw output for debug graph.")
 argparser.add_argument('--debug-graph-output-filename', help="The generated DOT graph filename. Defaults to 'dmake-services.debug.{group_by}.gv'")
 argparser.add_argument('--debug-graph-output-format', default='png', help="The generated DOT graph format (`png`, `svg`, `pdf`, ...).")
 
--- a/dmake/cli.py
+++ b/dmake/cli.py
@@ -38,7 +38,8 @@ argparser = argparse.ArgumentParser(prog='dmake')
 
 argparser.add_argument('--debug-graph', default=False, action='store_true', help="Generate dmake steps DOT graph for debug purposes.")
 argparser.add_argument('--debug-graph-and-exit', default=False, action='store_true', help="Generate dmake steps DOT graph for debug purposes then exit.")
-argparser.add_argument('--debug-graph-output-filename', default='dmake-services.debug.gv', help="The generated DOT graph filename.")
+argparser.add_argument('--debug-graph-group-by', default='command', choices=['command', 'height'], help="Group nodes by <> in generated DOT graph.")
+argparser.add_argument('--debug-graph-output-filename', help="The generated DOT graph filename. Defaults to 'dmake-services.debug.{group_by}.gv'")
 argparser.add_argument('--debug-graph-output-format', default='png', help="The generated DOT graph format (`png`, `svg`, `pdf`, ...).")
 
 subparsers = argparser.add_subparsers(dest='cmd', title='Commands')
--- a/dmake/commands/graph.py
+++ b/dmake/commands/graph.py
@@ -15,6 +15,8 @@ def entry_point(options):
         return "<{}<br/><i><font point-size='10'>{}</font></i>>".format(name, file)
 
     dot = Digraph(comment='DMake Services', filename=options.output, format=options.format)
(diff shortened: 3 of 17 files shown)


thomas-riccardi:
full graph, with new pretty visual formatting:
`DMAKE_PARALLEL_EXECUTION=1 CHANGE_BRANCH=master DMAKE_ON_BUILD_SERVER=1 dmake --debug-graph-and-exit --debug-graph-pretty --debug-graph-group-by height deploy \*`
dmake-services debug height gv

nl2412 (approved):

PR #474, opened 2020-12-23
Jenkins code generation: Cleanup lock, add parallel

thomas-riccardi:
Preparation work for jenkins parallel execution


maingoh (commented):
Can you explain a bit more how it works from the user point of view ? How do we set the max concurrency and what do you mean by "cmd not supported with bash"

PR #486, opened 2021-03-04
Fix bash invalid syntax error on empty stage

thomas-riccardi:
Fix regression introduced by 2d00408a8274ea3bc63d2a9de6142fc4481afe41

Bash doesn't like:
```
echo -e '
## Building App ##'
{
}
```
=> DMakefile: line 56: syntax error near unexpected token `}'

Solution: guarantee at lest one command is in the block: the stage name echo
```
{ echo -e '
## Building App ##'
}
```

### DIFF
--- a/dmake/core.py
+++ b/dmake/core.py
@@ -722,8 +722,7 @@ function ignoreErrors()
     for cmd, kwargs in cmds:
         if cmd == "stage":
             write_line("")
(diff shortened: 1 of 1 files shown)


Zanoellia (approved):
Tested locally with success

Good to go :heavy_check_mark:

PR #489, opened 2021-03-24
link_names are now checked to be unique

thomas-riccardi:
Why? Because previously 2 needed_services could have the same link_name with different definitions (env), and dmake would randomly use one or the other, especially since #472 which made the execution order non-deterministic.

Both across docker_links and needed_services.

needed_services multiple instances are allowed if they define the same
NeededServiceSerializer (same service_name, same env, etc..).

Fix tests:
- prefix all link_names with `test-`, like we did for service name
- update test-resources accordingly


dmake/deepobuild.py
```diff
@@ -1092,6 +1099,25 @@ def kind(self, kind):
...
+        LinkNames.link_names = dict()
+
+    @staticmethod
+    def check_duplicate_link_name(kind, link, file):
+        link_name = link.link_name
+        if not link_name:
+            return
+        if link_name in LinkNames.link_names:
+            other_kind, other_link, other_file = LinkNames.link_names[link_name]
+            if not (kind == other_kind and link == other_link):
```
maingoh, line 1117: I am not sure to understand why you need this condition, checking the name is not enough ? You recall this function multiple times potentially with the same link ? That's why you need this extra check ?

maingoh (approved):
LGTM !

dmake/deepobuild.py
thomas-riccardi: the goal is to support multiple needed_services identical: multiple services (A and B) may need to point to the same specialized service C. NeededServiceSerializer has its comparison operator/methods specially defined for that.
And for DockerLinkSerializer, the default comparison will return `not equal` for different instances, even if they have the same properties.

I'll add some documentation (and maybe test?) for that crucial point. It was only partially explained in the commit/PR description.

nl2412 (approved):
lgtm

dmake/deepobuild.py
```diff
@@ -1092,6 +1099,36 @@ def kind(self, kind):
...
+    def reset():
+        LinkNames.link_names = dict()
+
+    @staticmethod
+    def check_duplicate_link_name(kind, link, file):
+        """
+        Goal: guarantee that each different docker container has a unique link name, while allowing multiple services linking to the same docker container (thus without ambiguity)
+        Rules:
+        - docker_links: no duplicate link_name
+        - needed_services: can duplicate link_name iff NeededServices are equivalent (same service_name, same env_exports, but env_exports and other attributes can be different)
```
maingoh, line 1116:
```suggestion
        - needed_services: can duplicate link_name if NeededServices are equivalent (same service_name, same env_exports, but env_exports and other attributes can be different)
```

maingoh (approved):
Nice !

dmake/deepobuild.py
thomas-riccardi: it's https://en.wikipedia.org/wiki/If_and_only_if

maingoh: :O

PR #490, opened 2021-03-24
Remove old deprecated feature usage in tutorial/test: docker link internal env vars (used in entrypoints)

thomas-riccardi:
Should use env_export instead.

Not used elsewhere, hard to detect (names changed, so grep can't find them), and will be hard to maintain in the future.

Removed some entrypoints, replaced others by dev.entrypoints to keep testing it all.


nl2412 (approved):
lgtm

PR #493, opened 2021-04-01
Better message when an error occurred

thomas-riccardi:
- remove the possibility to skip cleanup: just keep dmake alive while
  error is investigated, then press enter to finish with cleanup
- add multiple suggestions on how to investigate the error
- better dmake run messages, with link to dmake stop


nl2412 (approved):
lgtm

PR #498, opened 2021-05-19
README: tell to use new shell after installation

Zanoellia:


thomas-riccardi (approved):
LGTM!

It seems https://github.com/Deepomatic/dmake/blob/d836ad2bc41ad3a214f59c34311e6472fd6e0b69/install.sh#L164 wasn't enough?

Zanoellia:
> LGTM!
> 
> It seems
> 
> https://github.com/Deepomatic/dmake/blob/d836ad2bc41ad3a214f59c34311e6472fd6e0b69/install.sh#L164
> 
> wasn't enough?

Not enough for the non yet awake myself of this morning 😂

PR #505, opened 2021-07-28
Add secrets to base image build

JulienFS:


dmake/deepobuild.py
```diff
@@ -227,80 +228,11 @@ def _serialize_(self, commands, path_dir):
...
-
-        if not self.raw_root_image:
-            # FIXME: copy key while #493 is not closed: https://github.com/docker/for-mac/issues/483
-            if common.key_file is not None:
-                common.run_shell_command('cp %s %s' % (common.key_file, os.path.join(tmp_dir, 'key')))
+        project_dir = os.path.join(common.root_dir, path_dir)
+        build_dir = os.path.join(project_dir, '.build')
+
+        # Make the build directory
+        common.run_shell_command('rm -rf %s; mkdir -p %s' % (build_dir, build_dir))
```
thomas-riccardi, line 235: We need to support multiple executions of dmake in the same directory at the same time.
That's why we use tmp_dir everywhere, let's keep using tmp_dir if possible.

it also creates difficulties with dockerignore files; if we really need to create new files in the repo dir, let's put them in `.dmake/`, it's how we do it to collect tests results among other things, (so it's usually excluded by dmake users).

dmake/deepobuild.py
```diff
@@ -318,21 +250,105 @@ def _serialize_(self, commands, path_dir):
...
+        # We compute the hash of the files in the order as they are provided
+        full_hash = hashlib.md5()
+        # Init the hash with a version string to prevent collision in case of future change
+        full_hash.update(b'v3')
+
+        # Create the Dockerfile
+        dockerfile = os.path.join(build_dir, 'Dockerfile')
+        with open(dockerfile, 'w') as f:
+            f.write('FROM %s@%s\n' % (self.root_image, root_image_digest))
+            f.write('WORKDIR base')
```
thomas-riccardi, line 262:
```suggestion
            f.write('WORKDIR /base')
```

dmake/deepobuild.py
```diff
@@ -318,21 +250,105 @@ def _serialize_(self, commands, path_dir):
...
                 raise e
 
+        # We compute the hash of the files in the order as they are provided
+        full_hash = hashlib.md5()
+        # Init the hash with a version string to prevent collision in case of future change
+        full_hash.update(b'v3')
+
+        # Create the Dockerfile
+        dockerfile = os.path.join(build_dir, 'Dockerfile')
+        with open(dockerfile, 'w') as f:
```
thomas-riccardi, line 260:
```suggestion
        with open(dockerfile, 'wb') as f:
```

thomas-riccardi (commented):
quick review comments

dmake/deepobuild.py
JulienFS: Thx for early review ! I was about to move to .dmake anyway since we cant build from two context and we eventually have to include some template files from dmake repo (and using / as context is obviously not an option).
Good point for the multiple executions, i'll make a unique tmp_dir inside .dmake

dmake/deepobuild.py
JulienFS: actually using `wb` would force me to .encode() each string. Should I do it though ?

dmake/utils/dmake_build_base_docker
```diff
@@ -125,37 +125,7 @@ fi
 if [ -z "${BASE_IMAGE_ID}" ]; then
     log "Docker base image not found in cache, building it (${BASE_IMAGE})"
 
-    if [ `uname` = "Darwin" ] && [ ! -z "${DMAKE_SSH_KEY}" ]; then
```
JulienFS, line 128: i'm getting rid of the whole SSH thing (as long as it doesn't break the global cache), lets not forget to clean up everywhere and mark the thing as broken

dmake/deepobuild.py
```diff
@@ -318,6 +318,28 @@ def _serialize_(self, commands, path_dir):
...
+        # Create the Dockerfile
+        # Note that by using a more fine-grained COPY and RUN we could better leverage local docker cache
+        # But to nicely leverage this we would have to edit the base install scripts, which would break the global cache
+        dockerfile = os.path.join(tmp_dir, 'Dockerfile')
+        with open(dockerfile, 'w') as f:
+            f.write('FROM %s@%s\n' % (self.root_image, root_image_digest))
+            f.write('COPY . /base_volume\n')
+            f.write('RUN /bin/bash /base_volume/make_base.sh\n')
+            f.write('RUN rm -rf /base_volume && mkdir /base_volume\n')
+            f.write('FROM scratch\n')
```
JulienFS, line 337: this effectively removes base_volume from the image, however the metadata of the image end up being very different from what it used to be (no more root image)

dmake/deepobuild.py
thomas-riccardi: ok, I'm traumatized from python2 issues, I didn't know it changed that much around open/fd in python3, let's keep it your way.

dmake/deepobuild.py
```diff
@@ -318,6 +319,46 @@ def _serialize_(self, commands, path_dir):
...
+
+            if not os.path.isabs(secret_path):
+                raise DMakeException("Invalid 'build_secrets': key '%s', path '%s': not an absolute path" % (secret_name, secret_path))
+
+            if not os.path.isfile(secret_path):
+                raise DMakeException("Invalid 'build_secrets': key '%s', path '%s': file not found" % (secret_name, secret_path))
+
+            # FIXME: docker build fails if there is a space or a comma in the secret_path
+            dockerfile_secrets_mounts += ' --mount=type=secret,id={} '.format(secret_name)
+            build_secrets_args += ' --secret id={},src={} '.format(secret_name, secret_path)
```
JulienFS, line 345: I tried to quote the id={},src={} in different ways but it always end up failing with spaces. There might be a problem with the way the docker cli forwards parameters to buildctl or something ?

thomas-riccardi: it's probably us/dmake who fail to properly propagate/escape non-trivial things across our generated `DMakefile` (sometimes bash, sometimes Jenkinsfile), calling `dmake_build_base_docker`, finally calling `docker build`)

thomas-riccardi: anyway, it's OK to limit ourselves to simple secret names

docs/FORMAT.md
```diff
@@ -44,14 +45,15 @@
                 - **python_requirements** *(file path, default = ``)*: Path to python requirements.txt.
                 - **python3_requirements** *(file path, default = ``)*: Path to python requirements.txt.
                 - **copy_files** *(array\<file or directory path\>, default = `[]`)*: Files to copy. Will be copied before scripts are ran. Paths need to be sub-paths to the build file to preserve MD5 sum-checking (which is used to decide if we need to re-build docker base image). A file 'foo/bar' will be copied in '/base/user/foo/bar'.
+                - **build_secrets** *(free style object, default = `{}`)*: Secrets to mount on /run/secrets/secret_name during build time.
```
JulienFS, line 48: We want to abstract the underlying buildkit options and use different configurations for the different `--mount` options (secrets, cache, build) since they accent different parameters.

JulienFS: ssh mount can be added in an other PR

dmake/utils/dmake_build_base_docker
```diff
@@ -37,7 +38,7 @@ DOCKER_IMAGE_TAG=$1; shift
 DOCKER_IMAGE_TAG_V1=$1; shift
 DMAKE_DIGEST=$1; shift
 PUSH_IMAGE=$1; shift
-
+BUILD_EXTRA_ARGS=("$@"); # We copy the whole array so we have an explicit variable name
```
JulienFS, line 41: @thomas-riccardi I fixed the problems about spaces, quotes and comma in secret paths. Basically I added quotes around the `src=/blabla` keyvalue pair to fix commas, escaped quotes in `/blabla`, passed each secret as a parameter to the build script then used a bash array as you suggested.

docs/FORMAT.md
JulienFS: @thomas-riccardi is it ok for you to do so ?

thomas-riccardi: maybe call that `mount_secrets` so we convey the fact it's a file that's mounted, and reuse the dockerfile buildkit vocabulary ("build run mount type secret").

dmake/utils/dmake_build_base_docker
thomas-riccardi:
```suggestion
BUILD_EXTRA_ARGS=("$@") # We copy the whole array so we have an explicit variable name
```

dmake/utils/dmake_build_base_docker
```diff
@@ -125,38 +126,13 @@ fi
...
-
-    CID=`cat ${TMP_DIR}/cid.txt`
-    if [ ! -z "${DMAKE_TMP_DIR}" ]; then
-        echo ${CID} >> ${DMAKE_TMP_DIR}/containers_to_remove.txt
+    cd ${TMP_DIR}
+    if [ ${#BUILD_EXTRA_ARGS[@]} -eq 0 ]; then
+      DOCKER_BUILDKIT=1 docker image build . --tag "${BASE_IMAGE}"
+    else
+      DOCKER_BUILDKIT=1 docker image build "${BUILD_EXTRA_ARGS}" . --tag "${BASE_IMAGE}"
     fi
```
thomas-riccardi, line 134:
```suggestion
    DOCKER_BUILDKIT=1 docker image build "${BUILD_EXTRA_ARGS[@]}" . --tag "${BASE_IMAGE}"
```
from quick tests it should work: `"${BUILD_EXTRA_ARGS[@]}"` should be expanded to zero arguments when empty array.
In any case, it should be needed anyway to transparently pass arguments to docker build. If it breaks things, we should ideally fix that elsewhere.

dmake/deepobuild.py
```diff
@@ -1018,7 +1061,7 @@ def get_mount_opt(self, service_name, dmake_file_path, env=None):
 
 class TestSerializer(YAML2PipelineSerializer):
     docker_links_names = FieldSerializer(deprecated="Use 'services:needed_links' instead", data_type="array", child = "string", migration='0001_docker_links_names_to_needed_links', default = [], example = ['mongo'], help_text = "The docker links names to bind to for this test. Must be declared at the root level of some dmake file of the app.")
-    data_volumes       = FieldSerializer("array", child = DataVolumeSerializer(), default = [], help_text = "The read only data volumes to mount. Only S3 is supported for now.")
```
thomas-riccardi, line 1021: let's fix DataVolumeSerializer.source doc too and remove S3-only mention there too.

thomas-riccardi (changes requested):
let's add an example usage in test/ project, it will help my review too

dmake/deepobuild.py
```diff
@@ -318,6 +319,48 @@ def _serialize_(self, commands, path_dir):
...
+
+            if not os.path.isabs(secret_path):
+                raise DMakeException("Invalid 'build_secrets': key '%s', path '%s': not an absolute path" % (secret_name, secret_path))
+
+            if not os.path.isfile(secret_path):
+                raise DMakeException("Invalid 'build_secrets': key '%s', path '%s': file not found" % (secret_name, common.wrap_cmd(secret_path)))
+
+            dockerfile_secrets_mounts += ' --mount=type=secret,id={} '.format(secret_name)
+            # We have to double quote the src={} since the secret parameter takes CSV format and commas in path would break the parsing
+            build_secrets_args.append('--secret=id={},"src={}"'.format(secret_name, secret_path))
```
JulienFS, line 347: There is no more black magic involved. Build secrets are passed as is down to the docker build command. Double quotes are considered literals all the way. The --secret arg exepects a CSV value that is then split around the '=' sign. The double quote is the way to escape commas in CSV.

dmake/deepobuild.py
JulienFS: `Only host path and s3 URLs are supported for now`
There is already a mention of the host path in the DataVolumeSerializer. We could remove the "only" word though.

dmake/utils/dmake_build_base_docker
JulienFS: SSH deployment should still be OK. Nothing to remove in the documentation since SSH git repo cloning in the base image was an undocumented feature.

dmake/deepobuild.py
```diff
@@ -144,7 +144,7 @@ def get_volume_id(self):
 
 
 class SharedVolumeMountSerializer(YAML2PipelineSerializer):
-    source = FieldSerializer("string", example = "datasets", help_text = "The shared volume name (declared in .")
+    source = FieldSerializer("string", example = "datasets", help_text = "The shared volume name (declared in root volumes).")
```
thomas-riccardi, line 147:
```suggestion
    source = FieldSerializer("string", example = "datasets", help_text = "The shared volume name (declared in root `volumes`).")
```

dmake/utils/dmake_build_base_docker
thomas-riccardi: SSH deployment is deprecated and unused anyway.

ssh access in base image is deprecated and unused for several years now.

=> all good

dmake/deepobuild.py
thomas-riccardi: double quote in path still doesn't work, and is hard to make it work because of the double escaping that is badly supported by our `common.wrap_cmd`. That's OK to not support these strange cases. maybe just say that in the comment?

JulienFS:
LGTM !

thomas-riccardi (approved):

PR #518, opened 2022-03-11
dmake release: cleanup messages, create draft release

thomas-riccardi:
also, don't quiet when git fetch tags: it hides errors.

TODO later: use https://docs.github.com/en/rest/reference/releases#generate-release-notes-content-for-a-release , with controllable previous_tag_name (it's not exposed from web ui)


oquerne (approved):
LGTM !

PR #532, opened 2022-06-28
Publish separate code coverage reports

thomas-riccardi:
Merging is not well adapted to periods of migrations between testtools (it seems to work well for separate projects/code source merges though).


thomas-riccardi:
cc @Zorrette : one report per xml file, cf https://build.deepomatic.com/job/dmake/job/PR-532/2/coverage/#coverageTree

S4ckBr (approved):

PR #541, opened 2022-10-31
Support docker image tag customization with environment variables

### LINKED ISSUE #526: Allow to personalize the built docker image tag
Today, one can specify a `tag` for a built image: https://github.com/Deepomatic/dmake/blob/cc614bf89e726511d186cee0c0f7be5dc329688b/dmake/docker_image.py#L105

But this is too limiting:
- If no tag is specified, the docker image tag by default is `[{:variant}-]{:branch_name}-{:build_id}`
- If a tag is specified, the docker image tag will be hardcoded and always the same.

In Deepomatic/thoth#761, we need to push two different images to the `deepomatic/thoth` Docker registry. But one cannot do it:
- if we put no tag in the docker field of the dmake field, both image names would be the same and would conflict
- if we put a tag in the docker field of the dmake field, subsequent release would have the same hardcoded tag and would conflict.

We would need to be able to use templates in the tag to customize the tag with the branch and build number.

(this is not an urgent need)

thomas-riccardi:
closes #526

Add special env vars prefixed with `_` just for this tag name specification, so we can easily reproduce the previous behavior: `${_BRANCH_SANITIZED_FOR_DOCKER}-${_BUILD_ID_OR_LATEST}${_VARIANT_SUFFIX}`
=> it's the new default.
=> Actual default generated value doesn't change.


maingoh (approved):
Goo

PR #563, opened 2024-04-08
Add image aliases

JulienFS:
Add images aliases with the nice side-effect that it allows to push images to multiple registries.


dmake/docker_image.py
```diff
@@ -128,20 +125,39 @@ def get_image_name(self, env=None, latest=False):
...
+        for image_name in image_names:
+            # When deploying, we need to push the image. We make sure that the image has a user
+            if len(image_name.split('/')) == 1:
+                image_name_without_tag = image_name.split(':')[0]
+                raise DMakeException("Service '{}' declares a docker image without a user name in config::docker_image::name so I cannot deploy it. I suggest to change it to 'your_company/{}'".format(service_name, image_name_without_tag))
+
+            check_private_flag = "1" if self.check_private else "0"
+            append_command(commands, 'sh', shell='dmake_push_docker_image "%s" "%s"' % (image_name, check_private_flag))
+            image_latest = self.get_image_name(env=env, latest=True)
+            append_command(commands, 'sh', shell='docker tag %s %s && dmake_push_docker_image "%s" "%s"' % (image_name, image_latest, image_latest, check_private_flag))
```
thomas-riccardi, line 160: the `latest` tag logic is wrongly implemented here: it tags the one and only canonical image_latest once from each `image_names`.

either we want to tag it only once, or we want a `latest` tag for all aliases too?

thomas-riccardi (changes requested):
`latest` tag logic bug

dmake/docker_image.py
JulienFS: thanks for catching  this one, I'm going to tag all images. My goal is to have the same behavior as for the main image.

PR #564, opened 2024-04-14
Fix escaped string for python3.12

vdel:
Using dmake with python 3.12 fails with:

```
dmake.common.ShellError: /Users/vincent/deepomatic/dmake/dmake/utils/dmake_replace_vars:36: SyntaxWarning: invalid escape sequence '\$'
  lvars = re.findall('\${([A-Za-z\-_]+)}', data)
```

This PR fixes it.


thomas-riccardi (approved):
LGTM

PR #569, opened 2024-07-21
Use supported coverage plugin

vdel:
The [Code Coverage Plugin](https://plugins.jenkins.io/code-coverage-api/) is deprecated and does not work anymore.

We follow the instruction of the linked page and replace `publishCoverage` by `recordCoverage`.

This has been tested on Jenkins, it leads to the apparition of 2 menu entries below "Test result", that were not present before this PR:
Capture d’écran 2024-07-21 à 20 56 54

The "Coverage report" looks like this:
Capture d’écran 2024-07-21 à 20 58 52

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -109,8 +109,7 @@ node {
           sh ". .venv3/bin/activate && pytest --numprocesses=4 -vv --color=yes --junit-xml=junit.xml --junit-prefix=python3 --cov=dmake/ --cov-report=xml:coverage.xml --cov-report html:cover/"
         } finally {
           junit keepLongStdio: true, testResults: 'junit.xml'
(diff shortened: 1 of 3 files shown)


thomas-riccardi (approved):
LGTM from afar;
- do we still need one unique record step globally like it was for CoberturaPublisher?
cf https://github.com/Deepomatic/dmake/pull/446/files#diff-56fb13b9ad11fb221716f5f3e08471322906429c18f0b352d38f142a199a48d4R581
- next step: deploy it in jenkins before merging here, and see how our other repos behave.
- later steps could be to publish results in github checks/comments

vdel:
> do we still need one unique record step globally like it was for CoberturaPublisher?

I do not know. I guess so, as it is likely to show only one entry in the left menu of the jobs.

> next step: deploy it in jenkins before merging here, and see how our other repos behave.

OK, should I do it? Directly in the Jenkins container ?

> later steps could be to publish results in github checks/comments

Indeed

Issue #445, opened 2020-04-30
Use new Jenkins Code Coverage API Plugin


thomas-riccardi:
See:
- https://plugins.jenkins.io/code-coverage-api/
- https://www.jenkins.io/doc/pipeline/steps/code-coverage-api/
- https://github.com/jenkinsci/code-coverage-api-plugin/wiki
- https://www.jenkins.io/blog/2018/08/17/code-coverage-api-plugin-1/

Why?
- more modern, code coverage unification
- It allows us to handle more formats than just cobertura (plugin-based).
- PR coverage diff
- more explorable coverage trend
- finer-grained reports grouping (we could do per dmake service, or per group of services with extra dmake.yml config, using plugin "tag" feature)

thomas-riccardi:
# v1
publish both on old and new plugin, no change to dmake.yml

# v2
- remove old coberturaPublisher
- add new `code_coverage` section in dmake.yml to replace `cobertura`
  - add options to control failure thresholds, reports grouping, etc..

thomas-riccardi:
coverage diff publish on github is not planned for now: all 2 plugins that do such things are reimplementing xml parsing instead of working on top of code-coverage-api-plugin...

thomas-riccardi:
skipped to v3: [Code Coverage Plugin](https://plugins.jenkins.io/code-coverage-api/) is deprecated, use now [Coverage Plugin](https://github.com/jenkinsci/coverage-plugin), cf #569

PR #574, opened 2025-01-07
Fix cobertura coverage collection

vdel:
Current behavior collects XML cobertura reports from container into a host directory randomly generated, which results in many directories in .dmake/cobertura_tests_results/, not all being from the same build.

For example:
```
jenkins@fce34241e9ef:~/jobs/vesta/branches/master/workspace$ ls -lt .dmake/cobertura_tests_results/
total 140
drwxr-xr-x 3 jenkins jenkins 4096 Jan  2 22:19 db6f4875-a58f-4793-bcda-0a21d6781ff0
...
drwxr-xr-x 3 jenkins jenkins 4096 Dec 26 09:56 8defe87f-9146-45c3-b073-f962de7000ad
```

When generating the report, we look for XML files in `dmake/cobertura_tests_results/**/*.xml`

### DIFF
--- a/dmake/core.py
+++ b/dmake/core.py
@@ -665,14 +665,8 @@ def generate_command_pipeline(file, cmds):
     indent_level += 1
 
     if emit_cobertura:
-        write_line("try {")
-        indent_level += 1
         write_line("recordCoverage(tools: [[parser: 'COBERTURA', pattern: '%s/**/*.xml']], qualityGates: [[metric: 'LINE', baseline: 'PROJECT_DELTA', criticality: 'NOTE']], sourceCodeRetention: 'NEVER')" % (cobertura_tests_results_dir))
         write_line('''sh('rm -rf "%s"')''' % cobertura_tests_results_dir)
-        indent_level -= 1
-        write_line("} catch (error) {")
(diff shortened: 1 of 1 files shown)


dmake/core.py
```diff
@@ -664,14 +665,8 @@ def write_line(data):
     indent_level += 1
 
     if emit_cobertura:
-        write_line("try {")
```
thomas-riccardi, line 667: why remove the try-catch? it was there to have a proper error message in case of missing coverage files, cf #487

thomas-riccardi (commented):
Quickfix LGTM, modulo removed-try-catch.

The real fix now that we use the new plugin, would be to do multiple calls to `recordCoverage`: one per service; cf https://github.com/Deepomatic/dmake/commit/412a63b7d1c33d9a379c220b400e84cf9cb4adbf#diff-f5d1a1c49ff4a7564e7e5542628e21c28cf23d141cee275805d942b54adcea67R449

we would then have the direct unique path for that service instance/run, cf `host_report`.

cf https://github.com/Deepomatic/dmake/issues/445#issuecomment-621704361

vdel:
Closing as the original issue is not well understood. Next step = more visibility on the coberura exception, and then a fix

